PR #40, opened 2019-12-12
Fixing DP Predicate Test and improving performance 1300X times

ankitufl:
Fixed the DP predicate test. It now correctly succeeds for all the DP equations and fails for the buggy implementation injected. More importantly, shortened its run time from 13 seconds for 10K iterations and 10K records to 0.01 seconds (1300X faster) for same iterations and same set of records. Did this by doing the exact aggregation once and adding vectorized noise 10K times instead of using loop for generating samples.


evaluation/Aggregation.py
```diff
@@ -4,36 +4,37 @@
 import numpy as np
 
 class Aggregation:
-    def __init__(self, epsilon=1.0, t = 1):
+    def __init__(self, epsilon=1.0, t = 1, repeatcount = 10000):
```
eedeleon: python convention has no spaces for named variables

t=1, repeat_count=10000 etc.

ankitufl: Removed spaces

ankitufl:
Finishing this merge as I have new changes in master branch to support evaluation of Burdock on same files and some new ones. Please review those set of commits.

PR #41, opened 2019-12-13
DP Evaluation on Burdock Private Query

ankit-oss:
Repeated querying of sample SQL query string using Burdock private query capability (parser, AST) on simulated neighboring datasets

Test is passing as shown in the Jupyter notebook.

It currently takes long to execute. We shall be optimizing this in future check-in by returning back a noisy array instead of single query response.


ankit-oss:
Thank you Josh!

Issue #24, opened 2019-12-02
Implement Bounded Multi-Histogram from Spec


joshua-oss:
Can implement with cross join and right outer join, or by caching distinct values and post-processing.  If post-processing, unsorted results should be shuffled

joshua-oss:
Will not implement

PR #57, opened 2019-12-22
DP Noise Bounds SDK Changes

ankit-oss:
1. Modified Private Query add noise function to also be able to output lower and upper bound rowsets
2. Used the upper and lower bound rowsets to compare with 10K noisy responses
3. Check if the noisy responses fall within the confidence level
4. Output count of responses within the error bounds and count which are outside

### DIFF
--- a/sdk/burdock/query/sql/private/query.py
+++ b/sdk/burdock/query/sql/private/query.py
@@ -42,7 +42,7 @@ class PrivateQuery:
         kcc = [kc for kc in subquery.keycount_symbols() if kc[0] != "keycount"]
         if len(kcc) > 0:
             srs["keycount"] = srs[kcc[0][0].lower()]
-        srs = srs.filter("keycount", ">", self.tau ** 2, srs.bounds)
+        srs = srs.filter("keycount", ">", self.tau ** 2)
 
         # add noise to all columns that need noise
         for nsym in subquery.numeric_symbols():
@@ -51,16 +51,12 @@ class PrivateQuery:
             sens = sym.sensitivity()
             mechanism = Laplace(self.epsilon, sens, self.tau)
             srs.bounds[name] = mechanism.bounds(pct)
-            if sym.type() == "int":
-                if sym.sensitivity() == 1:
-                    counts = mechanism.release(srs[name])
-                    counts[counts < 0] = 0
-                    srs[name] = counts
-                    # New rowset returned at below step. So passing bounds, else they get initialized again.
-                    srs = srs.filter(name, ">", self.tau, srs.bounds)
-                elif sens is not None:
-                    srs[name] = mechanism.release(srs[name])
-            elif sym.type() == "float" and sens is not None:
+            if sym.sensitivity() == 1:
+                counts = mechanism.release(srs[name])
+                counts[counts < 0] = 0
+                srs[name] = counts
+                srs = srs.filter(name, ">", self.tau)
+            else:
                 srs[name] = mechanism.release(srs[name])
 
         syms = query.all_symbols()
@@ -71,16 +67,12 @@ class PrivateQuery:
 
         srsc = srs.m_cols
         bindings = dict((name.lower(), srsc[name]) for name in srsc.keys())
-        #bound_bindings = dict((name.lower(), srs.bounds[name]) for name in srsc.keys())
 
         cols = []
-        bound_cols = []
         for c in query.select.namedExpressions:
             cols.append(c.expression.evaluate(bindings))
-            #bound_cols.append(c.expression.evaluate(bound_bindings))
         for idx in range(len(cols)):
             newrs[newrs.idxcol[idx]] = cols[idx]
-            #newrs.bounds[newrs.idxcol[idx]] = bound_cols[idx]
 
         # Now sort, if it has order by clause
         if query.order is not None:
--- a/sdk/burdock/query/sql/private/query.py
+++ b/sdk/burdock/query/sql/private/query.py
@@ -42,7 +42,7 @@ class PrivateQuery:
         kcc = [kc for kc in subquery.keycount_symbols() if kc[0] != "keycount"]
         if len(kcc) > 0:
             srs["keycount"] = srs[kcc[0][0].lower()]
-        srs = srs.filter("keycount", ">", self.tau ** 2)
+        srs = srs.filter("keycount", ">", self.tau ** 2, srs.bounds)
 
         # add noise to all columns that need noise
         for nsym in subquery.numeric_symbols():
@@ -56,7 +56,8 @@ class PrivateQuery:
                     counts = mechanism.release(srs[name])
                     counts[counts < 0] = 0
                     srs[name] = counts
-                    srs = srs.filter(name, ">", self.tau)
+                    # New rowset returned at below step. So passing bounds, else they get initialized again.
+                    srs = srs.filter(name, ">", self.tau, srs.bounds)
                 elif sens is not None:
                     srs[name] = mechanism.release(srs[name])
             elif sym.type() == "float" and sens is not None:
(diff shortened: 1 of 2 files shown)


sdk/burdock/query/sql/private/query.py
```diff
@@ -52,17 +53,40 @@ def _apply_noise(self, subquery, query, syms, types, sens, srs):
...
+                        lower_bound = srs[name] + self.mechanism.bounds(pct)[0]
+                        srs[name] = lower_bound
+                    elif(bounds == "upper"):
+                        upper_bound = srs[name] + self.mechanism.bounds(pct)[1]
+                        srs[name] = upper_bound
+                    else:
+                        counts = self.mechanism.release(srs[name])
+                        counts[counts < 0] = 0
+                        srs[name] = counts
+                        srs = srs.filter(name, ">", self.tau)
```
joshua-oss: This line is probably unnecessary, but defer fix to issue #24

sdk/burdock/query/sql/private/query.py
```diff
@@ -52,17 +53,40 @@ def _apply_noise(self, subquery, query, syms, types, sens, srs):
             name, sym = nsym
             name = name.lower()
             sens = sym.sensitivity()
+            
             if sym.type() == "int":
```
joshua-oss: Does typed recordset coerce these to ints, or should we do it here?  The new release() function doesn't return ints

sdk/burdock/query/sql/private/query.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, reader, metadata, epsilon=4.0):
         self.tau = 5
 
         #self.mechanism = Gaussian(epsilon, 100000, 1, self.tau)
-        self.mechanism = Laplace(self.epsilon, self.tau)
+        self.mechanism = Laplace(self.epsilon, tau=self.tau)
```
joshua-oss: I think we can just delete this and construct the mechanism where needed below, since the sensitivity is tied to each instance now.

sdk/burdock/query/sql/private/query.py
```diff
@@ -52,17 +53,40 @@ def _apply_noise(self, subquery, query, syms, types, sens, srs):
...
-                srs[name] = self.mechanism.sum_float(srs[name], sens)
-
+                mechanism = Laplace(self.epsilon, sens, self.tau)
+                if(bounds == "lower"):
+                    lower_bound = srs[name] + mechanism.bounds(pct)[0]
+                    srs[name] = lower_bound
+                elif(bounds == "upper"):
+                    upper_bound = srs[name] + mechanism.bounds(pct)[1]
+                    srs[name] = upper_bound
+                else:
```
joshua-oss: I'm not sure what this is doing.  It looks like it is returning bounds centered on the exact value for each row of the histogram, which means the midpoint would reveal the exact value.  And bounds should change based on the value of the cell, so bounds should be returned centered on 0, and only once per column.  Also, this could all be done in a single pass.  So, for example:

srs.bounds[name] = mechanism.bounds(pct)
srs[name] = mechanism.release(...

The bounds dict would contain 2-tuples (lower, upper) with the zero-centered bounds that apply to the whole column.

evaluation/Aggregation.py
```diff
@@ -1,60 +1,62 @@
...
+    # Run the query using the private reader and input query
+    # Get query response back
+    def run_agg_query(self, df, metadata_path, query, confidence):
+        metadata = MetadataLoader(metadata_path).read_schema()
+        reader = CSVReader(metadata, df)
+        private_reader = PrivateQuery(reader, metadata, self.epsilon)
+        exact_values = private_reader._execute_exact(query)
+        lower_bound = private_reader._apply_noise_bounds(*exact_values, bounds = "lower", pct = confidence)[1:][0][0]
+        upper_bound = private_reader._apply_noise_bounds(*exact_values, bounds = "upper", pct = confidence)[1:][0][0]
+        return np.array([private_reader._apply_noise_bounds(*exact_values)[1:][0][0] for i in range(self.repeat_count)]), lower_bound, upper_bound
```
joshua-oss: See below, this can be a single call

sdk/burdock/query/sql/private/query.py
ankitufl: Deleted

sdk/burdock/query/sql/private/query.py
ankitufl: Typed rowset does the coercing in line #99 of rowset.py. So we need not do it here. 
self.m_cols[cn] = np.array(value).astype(int)

sdk/burdock/query/sql/private/query.py
ankitufl: Centered bounds on 0 and stored them in srs.bounds to be retrieved in single pass.

sdk/burdock/query/sql/private/query.py
ankitufl: I see. Sure, keeping it as-is then.

sdk/burdock/query/sql/private/query.py
```diff
@@ -40,29 +37,31 @@ def execute(self, query_string):
...
         return self._apply_noise(*exact_values)
 
-    def _apply_noise(self, subquery, query, syms, types, sens, srs):
+    def _apply_noise(self, subquery, query, syms, types, sens, srs, pct=0.95):
         # if user has selected keycount for outer query, use that instead
         kcc = [kc for kc in subquery.keycount_symbols() if kc[0] != "keycount"]
         if len(kcc) > 0:
             srs["keycount"] = srs[kcc[0][0].lower()]
-        srs = srs.filter("keycount", ">", self.tau ** 2)
+        srs = srs.filter("keycount", ">", self.tau ** 2, srs.bounds)
```
joshua-oss: This can be done in typed recordset

sdk/burdock/query/sql/private/query.py
```diff
@@ -40,29 +37,31 @@ def execute(self, query_string):
...
 
         # add noise to all columns that need noise
         for nsym in subquery.numeric_symbols():
             name, sym = nsym
             name = name.lower()
             sens = sym.sensitivity()
+            mechanism = Laplace(self.epsilon, sens, self.tau)
+            srs.bounds[name] = mechanism.bounds(pct)
```
joshua-oss: Can collapse out sets and type check

sdk/burdock/query/sql/private/query.py
```diff
@@ -94,9 +97,8 @@ def _apply_noise(self, subquery, query, syms, types, sens, srs):
             sf = [("-" if desc else "") + colname for colname, desc in sort_fields]
 
             newrs.sort(sf)
-
-        return newrs.rows()
         
+        return (newrs.rows(), srs.bounds)
```
joshua-oss: For example: SELECT AVG(Age)

Step 0-2:  SELECT COUNT(Age) AS ca, SUM(Age) AS sa. # have bounds of dimension 2 (srs)
Step 3: SELECT sa / ca AS aa # output is dimension 1, no bounds (csrs)

~+~+~+~+~+~+~~+

Step 0: SELECT SUM(Age)
Step 3: SELECT 5 + SUM(Age) AS a5, 5 * SUM(Age), SUM(5 * Age) AS at5

CountA=1 (very noisy, close to 0), SumA=10;;  CountB=999 (less noise), SumB=9990 === AVGA = 10.0, AVGB == 10.0

sdk/burdock/query/sql/private/query.py
ankit-oss: Collapsed type checks. Retained check of COUNT sensitivity as it does clamping of COUNT less than 0 to 0 and tau filtering.

PR #68, opened 2020-01-07
Wrote code to create new example dataset, overwrote example.csv

mayigrin:
Will use this new dataset to test IBM diff priv linear & logistic regression models


mayigrin:
@eedeleon I know it's a really simple PR, but we were having trouble figuring out how to share code between the two of us directly, and we want to distribute this dataset so that we can both work with it.

PR #69, opened 2020-01-08
Made MLFlow project containing IBM's diff. priv. linear regression

mayigrin:
I had to add certain parameters for the MLProject, namely `x_features` which takes in a list of the desired features' columns' (i.e. X) names, and `y_targets` which takes in a list of the desired targets' columns' (i.e. y) names.

It uses the package `diffprivlib` to access the IBM model. One other thing of note is that IBM's differentially private linear regression model asks for X ranges and y ranges, which we can calculate given the user's X and y definitions and the schema-specified max&min values. However, it also asks for the data norm. Since that information isn't currently provided in the schema, my code currently calculates it and passes that value to the model. In the future though, that should be changed in order to preserve differential privacy. I use `logging.warning()` to warn the user about this when the python file is run.


service/modules/lin-reg-module/README.md
```diff
@@ -0,0 +1 @@
+# Partial Differentiall Private Linear Regression module Mlflow project
```
esfahany: Typo; differentially

service/modules/lin-reg-module/conda.yaml
```diff
@@ -0,0 +1,9 @@
+name: tutorial
+channels:
+  - defaults
+dependencies:
+  - pip:
+    - mlflow
+    - pandas
+    - diffprivlib
+    - git+https://git.repo/some_repo.git#egg=version_subpkg&subdirectory=repo
```
eedeleon: This bug is inherited but we should probably resolve it before it spreads further. 

The line should be 

pip install git+https://github.com/privacytoolsproject/burdock.git#subdirectory=sdk


The line does not currently get called since most module executions remove the conda setup step

service/modules/lin-reg-module/conda.yaml
```diff
@@ -0,0 +1,9 @@
+name: tutorial
+channels:
+  - defaults
+dependencies:
```
esfahany: numpy?

mayigrin: Don't think it's necessary based on sql-module, as per our conversation

service/modules/lin-reg-module/linreg.py
```diff
@@ -0,0 +1,58 @@
...
+import mlflow.sklearn
+import json
+import sys
+import logging
+
+import numpy as np
+import pandas as pd
+
+from burdock.client import get_dataset_client
+from burdock.data.adapters import load_reader, load_metadata, load_dataset
```
eedeleon: no need for the load_reader

service/modules/lin-reg-module/linreg.py
```diff
@@ -0,0 +1,58 @@
...
+        #  Eventually, this will need to be changed, as it is bad practice
+        norms = np.linalg.norm(dataset, axis=1)
+        max_norm = np.amax(norms)
+        logging.warning('Currently calculating the data norm instead of using schema-specified value. \
+                        This is bad practice, and will eventually be changed')
+
+        model = LinearRegression(data_norm=max_norm, epsilon=budget, range_X=x_range, range_y=y_range).fit(X, y)
+
+        # Log mlflow attributes for mlflow UI
+        mlflow.log_param("dataset_name", dataset_name)
```
eedeleon: the params should be logged at the start of the run. It helps keep more info even when things fail

service/modules/lin-reg-module/linreg.py
```diff
@@ -0,0 +1,58 @@
...
+        mlflow.log_param("budget", budget)
+        mlflow.log_param("x_features", x_features)
+        mlflow.log_param("y_targets", y_targets)
+        mlflow.sklearn.log_model(model, "model")
```
eedeleon: we should log a result.json like we do in the sql module

the spec could be 
{
run_id: mlflow.active_run().info.run_id,
model_name: model_name
}

PR #70, opened 2020-01-09
Log reg module

esfahany:
Implemented IBM's differentially private logistic regression. Modeled after Maya's linear regression, similar changes were made and will need to be made to various parts of this module. Parameters were added to allow users to specify feature/target columns of the dataset. A warning was added for data norm, which is currently not provided in the schema and is directly calculated. This will be changed in the future to preserve differential privacy.


service/modules/logreg-module/conda.yaml
```diff
@@ -0,0 +1,10 @@
+name: tutorial
+channels:
+  - defaults
+dependencies:
+  - pip:
+    - mlflow
+    - pandasql
```
mayigrin: I don't know if `pandasql` is necessary here? I don't think you're making any sql queries so I'm not sure. Something worth looking into.

service/datasets/example.yaml
```diff
@@ -2,16 +2,16 @@ Database:
...
         type: int
         min: 0
-        max: 25
-      C:
+        max: 20
+      c:
         type: int
         min: 0
-        max: 125
+        max: 1
```
mayigrin: Pretty sure these conflicts with `externs_ml` will go away when we push my PR, so let's just keep in mind to do that first and I think we won't have to worry about them.

service/modules/logreg-module/conda.yaml
eedeleon: It is not. We use it for the sql module

service/datasets/create_example_dataset.py
```diff
@@ -0,0 +1,20 @@
+import random
+
+"""
+Code to create an example dataset for testing. c is y-value; 
+c is 0 if a is low and is high, and is 1 otherwise
+"""
+
+list_a = [random.randint(0, 50) for i in range(1000)]
+list_b = [random.randint(0, 50) for i in range(1000)]
+f = open("example.csv", "w", newline='')
```
eedeleon: Switching the f = + f.close to 

with open("example.csv", "w", newline='') as f:
     the rest of the code

Helps ensure close is called in the case of a failure. It uses the context manager paradigm in python

service/modules/logreg-module/run_query.py
```diff
@@ -0,0 +1,47 @@
+import mlflow
```
eedeleon: We should rename run_query.py

service/modules/logreg-module/run_query.py
```diff
@@ -0,0 +1,47 @@
...
+import mlflow.sklearn
+import json
+import sys
+import logging
+
+import numpy as np
+import pandas as pd
+
+from burdock.client import get_dataset_client
+from burdock.data.adapters import load_reader, load_metadata, load_dataset
```
eedeleon: no need for reader

service/modules/logreg-module/run_query.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # TODO change how we find data norm (bad practice here in calculating the true norm; will be changed to use one specified in the schema)
+        norms = np.linalg.norm(dataset, axis=1) # norms for each column
+        max_norm = np.amax(norms)
+        logging.warning('Currently calculating the data norm instead of using schema-specified value. This is bad practice, and will eventually be changed') # provide a warning about bad practice
+
+        model = LogisticRegression(data_norm=max_norm, epsilon=budget).fit(X, y)
+
+        # log attributes
+        mlflow.log_param("dataset_name", dataset_name)
```
eedeleon: we should move these to right after the with mlflow line

service/modules/logreg-module/run_query.py
```diff
@@ -0,0 +1,47 @@
...
+        mlflow.log_param("budget", budget)
+        mlflow.log_param("x_features", x_features)
+        mlflow.log_param("y_targets", y_targets)
+        mlflow.sklearn.log_model(model, "model")
```
eedeleon: We need to add the results.json line. I commented on the other PR with the spec and guidelines.

service/modules/logreg-module/MLproject
```diff
@@ -0,0 +1,12 @@
...
+conda_env: conda.yaml
+
+entry_points:
+  main:
+    parameters:
+      dataset_name: str
+      budget: float
+      x_features: str
+      y_targets: str
+    command: "python run_query.py {dataset_name} {budget} {x_features} {y_targets}"
```
eedeleon: this will need to be changed when you change the file name

service/modules/logreg-module/MLproject
```diff
@@ -0,0 +1,12 @@
+name: diffpriv_sql
```
eedeleon: this should be updated

service/modules/logreg-module/conda.yaml
esfahany: Thanks for catching it!

service/modules/logreg-module/run_query.py
esfahany: removed the reader

service/datasets/example.yaml
esfahany: Sounds good.

service/datasets/example.csv
```diff
@@ -1,2 +1,1001 @@
 a,b,c
```
mayigrin: Edi just mentioned to me that it's better not to upload data to the GitHub, so you might want to delete this file from the remote.

PR #71, opened 2020-01-10
DP Powerset Test and Pytest stochastic tests

ankit-oss:
Currently, DP Predicate test ran by removing a random record. In this PR, we implement the loop above DP predicate test that generates 12 neighbor pairs by exhaustively doing DFS on database search graph of size 3 i.e. 3 records in database. 

Powerset currently runs for 1 database of size 3. In next check-in, we shall run it for many such databases of size 3 diversely generated via Halton sequences. 

We uncovered some issues while running this test - 

1. NULL values returned by SDK intermittently on a dataset of size 3
2. SUM or COUNT querying not supported on empty dataset

We workaround first issue by filling null values with previous noisy response or actual value if previous noisy non-null response does not exist. 
We workaround second issue by not running the powertest on empty databases

Jupyter notebook has the results of the DP predicate test applied on the powerset.


joshua-oss:
FYI, this appears to be a PR from your personal GitHub to your feature branch in burdock.  I am approving everything in this PR, since it will go into your feature branch.  Then I will pull down the feature branch and test.

PR #72, opened 2020-01-10
Bounds expr

joshua-oss:
Small updates to move bounds off of the OSBC-compat execute method

### DIFF
--- a/sdk/burdock/query/sql/metadata/release.py
+++ b/sdk/burdock/query/sql/metadata/release.py
@@ -1,0 +1,30 @@
+"""
+Represents the metadata for a differentially private release.
+
+We currently implement a subset of PSI release metadata.  This
+will be updated to align with yarrow after yarrow enhanced release
+metadata format is specified.
+"""
+class Result:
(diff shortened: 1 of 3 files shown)


sdk/burdock/query/sql/metadata/release.py
```diff
@@ -0,0 +1,28 @@
...
+        :param float delta: The delta privacy parameter used
+        :param string[] release: Vector of noisy release values as string
+        :param ((float, float, float)[]) intervals: Lower and upper bounds of confidence intervals at each pct.
+         By default, only 95% is returned, but caller can request multiple.  The interval
+         returned is centered on 0, and should not be thought of as an error bound.
+        """
+        self.statistic = statistic
+        self.mechanism = mechanism
+        self.name = name
+        self.epsilon = epsilon
```
eedeleon: We will need a .serialize function for saving it to disk

eedeleon: We can add it after

Issue #75, opened 2020-01-10
Private Query exception when doing SUM or COUNT query on empty database


ankit-oss:
**Repro Steps**
ag = agg.Aggregation(t=1, repeat_count=10000)
query_str = "SELECT SUM(Usage) AS TotalUsage FROM "
res_list = self.dp_powerset_test(query_str, plot=False)

Within powerset_test, we can invoke below commands and create empty datasets [0_0, 1_1, 2_2] - 

        ag = agg.Aggregation(t=1, repeat_count=repeat_count)
        ex = exp.Exploration()
        ex.generate_powerset(ex.df)
        res_list = {}
        for filename in ex.visited:
            if(filename not in ["0_0", "1_1", "2_2"]):
                d1_query = query_str + "d1_" + filename + "." + "d1_" + filename
                d2_query = query_str + "d2_" + filename + "." + "d2_" + filename
                d1 = pd.read_csv(os.path.join(ex.file_dir, ex.csv_path , "d1_" + filename + ".csv"))
                d2 = pd.read_csv(os.path.join(ex.file_dir, ex.csv_path , "d2_" + filename + ".csv"))
                d1_yaml_path = os.path.join(ex.file_dir, ex.csv_path , "d1_" + filename + ".yaml")
                d2_yaml_path = os.path.join(ex.file_dir, ex.csv_path , "d2_" + filename + ".yaml")
                fD1, fD1_bounds = ag.run_agg_query(d1, d1_yaml_path, d1_query, confidence)
                fD2, fD2_bounds = ag.run_agg_query(d2, d2_yaml_path, d2_query, confidence)

0_0 means D1 has 1 row of 0 index and D2 has that dropped. So D2 is empty database. run_agg_query crashes on this line - 

fD2, fD2_bounds = ag.run_agg_query(d2, d2_yaml_path, d2_query, confidence)

**Error**
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-19-eb648738bff6> in <module>
      1 start_time = time.time()
      2 query_str = "SELECT COUNT(UserId) AS TotalUsage FROM "
----> 3 res_list = dv.dp_powerset_test(query_str, plot=False)
      4 print("DP Powerset Test: ", res_list)
      5 print("--- %s seconds ---" % (time.time() - start_time))

/e/Git/burdock/evaluation/DPVerification.py in dp_powerset_test(self, query_str, debug, plot, bound, exact, repeat_count, confidence)
[...]

Issue #74, opened 2020-01-10
Intermittent NULL responses while repeated SUM or COUNT querying


ankit-oss:
When doing a repeated query 10K times on a dataset of 3 records and running a COUNT or SUM query, there are empty results retured intermittently. This leads DP predicate test to not be able to work reliably. We have a workaround in place to either fill in using the previous value or fill using actual value if previous non-null noisy value not available. 

**Repro**
1. Run the main method [here](https://github.com/ankit-oss/burdock/blob/master/evaluation/DPVerification.py) and set breakpoint on it

ag = agg.Aggregation(t=1, repeat_count=10000)
query_str = "SELECT SUM(Usage) AS TotalUsage FROM "
res_list = self.dp_powerset_test(query_str, plot=False)

2. Watch noisy_values returned [here](https://github.com/ankit-oss/burdock/blob/0eb6414ba687a9363f089d795cda668753f8babc/evaluation/Aggregation.py#L63) for first dataset of size 3

noisy_values = [private_reader._apply_noise(*exact_values)[0][1:] for i in range(self.repeat_count)]

**Output Results**
>0000:[(19892,)]
>0001:[(-10639,)]
>0002:[(31980,)]
>0003:[(8676,)]
>0004:[(14947,)]
>0005:[]
>0006:[(17043,)]
>0007:[(17194,)]
>0007:[]

[Jupyter notebook](https://github.com/ankit-oss/burdock/blob/master/evaluation/Differential%20Privacy%20Verification.ipynb) runs the Powerset test and finds in case of 3 records, empty list like in >0005 or >0007 are returned 3298 out of 10000 times

Count of times noisy values NULL:  3298  out of  10000  times

The NULL responses increases on database of 2 records - 
Count of times noisy values NULL:  7466  out of  10000  times

ankit-oss:
Thanks for the fix Josh! I pulled changes from the no_tau branch and don't see this intermittent issue anymore. So, closing this.

PR #78, opened 2020-01-12
Fix intermittent error on very small dataset

joshua-oss:
This is the two-line bug fix for bug #74 

Added 3 unit tests to verify

### DIFF
--- a/sdk/burdock/query/sql/private/query.py
+++ b/sdk/burdock/query/sql/private/query.py
@@ -108,7 +108,8 @@ class PrivateQuery:
                 counts[counts < 0] = 0
                 srs[name] = counts
 
-        srs = srs.filter("keycount", ">", self.tau ** 2)
+        if subquery.agg is not None:
+            srs = srs.filter("keycount", ">", self.tau ** 2)
 
         syms = query.all_symbols()
         types = [s[1].type() for s in syms]
@@ -149,7 +150,7 @@ class PrivateQuery:
         """
         if isinstance(query, str):
             raise ValueError("Please pass ASTs to execute_typed.  To execute strings, use execute.")
-        trs = self.execute_ast_exact(query)
+        trs = self.execute_ast_typed(query)
         return trs.rows()
 
     def execute_ast_typed(self, query):
--- a/sdk/burdock/query/sql/private/query.py
+++ b/sdk/burdock/query/sql/private/query.py
@@ -99,6 +99,8 @@ class PrivateQuery:
             name, sym = nsym
             name = name.lower()
             sens = sym.sensitivity()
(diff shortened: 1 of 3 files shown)


joshua-oss:
FYI, spec has also been updated to document that queries with no grouping expression do not apply tau thresholding:

https://docs.google.com/document/d/1Q4lUKyEu2W9qQKq6A0dbo0dohgSUxitbdGhX97sUNOM/edit#

joshua-oss:
Also fixes #75, pending confirmation from Ankit

sdk/burdock/query/sql/private/query.py
```diff
@@ -149,7 +152,7 @@ def execute_ast(self, query):
         """
         if isinstance(query, str):
             raise ValueError("Please pass ASTs to execute_typed.  To execute strings, use execute.")
-        trs = self.execute_ast_exact(query)
```
eedeleon: this seemed like a bigger bug than before. Did the stochastic tester catch this/

eedeleon: Is introducing 0 as the base value valid for DP? It is implied to be low anyways with the null return I suppose

tests/query/test_query.py
```diff
@@ -54,3 +55,46 @@ def test_group_by_noisy_typed_order_desc(self):
...
         rs = private_reader.execute_typed("SELECT COUNT(*) AS c, married AS m FROM PUMS.PUMS GROUP BY married ORDER BY c DESC")
         assert(rs['c'][0] > rs['c'][1])
+    def test_no_tau(self):
+        # should never drop rows
+        reader = CSVReader(schema, df)
+        private_reader = PrivateQuery(reader, schema, 4.0)
+        for i in range(10):
+            rs = private_reader.execute_typed("SELECT COUNT(*) AS c FROM PUMS.PUMS WHERE age > 90 AND educ = '8'")
+            assert(len(rs['c']) == 1)
+    def test_no_tau_noisy(self):
```
eedeleon: There can be some code sharing here but for now we can leave as is.

I can look into fixtures and code sharing for similar queries and tests.

Also, it is good to avoid prints during tests since they can add up to a noisy log file. logging.getLogger(__name__).debug("message") helps with this since it can be configured to show or not

sdk/burdock/query/sql/private/query.py
joshua-oss: That bug wasn't getting hit because there were no tests for it.  The other bug requiring setting NULL == 0 was indeed caught by the stochastic tester.  Setting null equal to 0 is valid (and required) in the case of counts, and I *think* in the case of other statistics, but I will need to add more tests

PR #79, opened 2020-01-12
Unit tests for empty query results

joshua-oss:
No bugs found, just adding unit tests

### DIFF
--- a/sdk/burdock/query/sql/private/query.py
+++ b/sdk/burdock/query/sql/private/query.py
@@ -99,6 +99,8 @@ class PrivateQuery:
             name, sym = nsym
             name = name.lower()
             sens = sym.sensitivity()
+            # treat null as 0 before adding noise
+            srs[name] = np.array([v if v is not None else 0.0 for v in srs[name]])
             mechanism = Laplace(self.epsilon, sens, self.tau)
             srs.bounds[name] = mechanism.bounds(pct)
             srs[name] = mechanism.release(srs[name])
--- a/sdk/burdock/query/sql/reader/rowset.py
+++ b/sdk/burdock/query/sql/reader/rowset.py
@@ -92,15 +92,15 @@ class TypedRowset:
             cn = key
             t = self.types[cn]
             if t == "string":
-                self.m_cols[cn] = np.array(value).astype(str)
+                self.m_cols[cn] = np.array([str(v) if v is not None else v for v in value])
             elif t == "boolean":
-                self.m_cols[cn] = np.array(value).astype(bool)
+                self.m_cols[cn] = np.array([bool(v) if v is not None else v for v in value])
             elif t == "int":
-                self.m_cols[cn] = np.array(value).astype(int)
+                self.m_cols[cn] = np.array([int(v) if v is not None else v for v in value])
             elif t == "float":
-                self.m_cols[cn] = np.array(value).astype(float)
+                self.m_cols[cn] = np.array([float(v) if v is not None else v for v in value])
             elif t == "datetime":
-                self.m_cols[cn] = [datetime.datetime(v) for v in value]
+                self.m_cols[cn] = np.array([datetime.datetime(v) if v is not None else None for v in value])
             else:
                 raise ValueError("Trying to load unknown type " + t)
         elif type(key) is int:
--- a/tests/query/test_query.py
+++ b/tests/query/test_query.py
@@ -3,6 +3,7 @@ import pandas as pd
 
 from burdock.query.sql.reader import CSVReader
 from burdock.query.sql import MetadataLoader
+from burdock.query.sql import QueryParser
 from burdock.query.sql.private.query import PrivateQuery
 from burdock.query.sql.reader.rowset import TypedRowset
 from pandasql import sqldf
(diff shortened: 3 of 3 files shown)


evaluation/Aggregation.py
```diff
@@ -56,9 +56,12 @@ def run_agg_query(self, df, metadata_path, query, confidence):
         metadata = MetadataLoader(metadata_path).read_schema()
```
eedeleon: the files with .py should be snake cased to follow python convention. It is not introduced here but evaluation/aggregation.py would be preferred. 

FYI In case a rename is attempted. It is sometimes non trivial to rename a file if only case sensitivity is changed due to file system nuances.

evaluation/Aggregation.py
```diff
@@ -56,9 +56,12 @@ def run_agg_query(self, df, metadata_path, query, confidence):
...
-        exact_values = private_reader._execute_exact(query)
-        bounds_centered_zero = list(private_reader._apply_noise(*exact_values, confidence)[1].values())[1]
-        actual_value = exact_values[-1].rows()[1:][0][1]
-        bounds = np.array([bounds_centered_zero[0] + actual_value, bounds_centered_zero[1] + actual_value])
-        noisy_values = np.array([private_reader._apply_noise(*exact_values)[0][1:][0][0] for i in range(self.repeat_count)])
-        return noisy_values, bounds
+        query_ast = private_reader.parse_query_string(query)
+        subquery_results = private_reader._preprocess(query_ast)
+
+        #exact_values = private_reader.execute_ast(query)
```
eedeleon: can we delete the commented code?

evaluation/DPVerification.py
```diff
@@ -5,12 +5,15 @@
 # i.e. passing (epsilon, delta) - DP condition
 # If the definition is not passed, there is a bug or it is a by-design bug in case of passing actual aggregates
 
+import sys
+import os
+sys.path.append(os.path.join(os.path.dirname(__file__), '..'))
```
eedeleon: Why is this sys.path line needed? Should some of this code live in the burdock library to avoid this?

evaluation/DPVerification.py
```diff
@@ -5,12 +5,15 @@
...
 import pandas as pd
 import numpy as np
 import math
 import matplotlib.pyplot as plt
-import Aggregation as agg
-import os
+import evaluation.Aggregation as agg
+import evaluation.Exploration as exp
```
eedeleon: same naming convention comment from before for exploration.py

tests/evaluation/test_stochastic.py
```diff
@@ -0,0 +1,66 @@
...
+        d2_query = "SELECT SUM(Usage) AS TotalUsage FROM d2.d2"
+        dp_res, acc_res = dv.dp_query_test(d1_query, d2_query, plot=False, repeat_count=10000)
+        print("Result of DP Predicate Test on SUM Query: ", dp_res)
+        assert(dp_res == True)    
+    
+    def test_dp_predicate_mean(self):
+        d1_query = "SELECT AVG(Usage) AS MeanUsage FROM d1.d1"
+        d2_query = "SELECT AVG(Usage) AS MeanUsage FROM d2.d2"
+        dp_res, acc_res = dv.dp_query_test(d1_query, d2_query, plot=False, repeat_count=10000)
+        print("Result of DP Predicate Test on MEAN Query: ", dp_res)
```
eedeleon: we should log these results with the logging library

import logging
test_logger = logging.getLogger(__name__)  # at the top of the file
logging.setLogLevel(logging.DEBUG)  # while writing/debugging the tests

test_logger.debug("message")  # in the tests.

evaluation/Aggregation.py
joshua-oss: None of these were supposed to be part of this PR, I'm not sure what happened

ankit-oss:
These are some good review comments in the evaluator piece. I'll use these to make changes in the its feature branch - mechanism_test and send a PR to push it into master.

evaluation/Aggregation.py
ankit-oss: This helped me rename - 
git mv previous_file_name new_file_name

evaluation/Aggregation.py
ankit-oss: Deleted

evaluation/DPVerification.py
ankit-oss: evaluation classes are not in burdock and they're being called from the Jupyter notebook as well as the pytest in tests/evaluation folder. This is helping support both right now. May be there's a more elegant way to do this. If I remove this, I get a ModuleNotFound error, even though I have added __init__.py in evaluation folder.

evaluation/DPVerification.py
ankit-oss: Renamed to exploration.py instead of Exploration.py

tests/evaluation/test_stochastic.py
ankit-oss: Nice. Thanks for the snippet. Used it for logging prints.

evaluation/DPVerification.py
joshua-oss: We can move to its own separate module in the future.  In the short term, you can run local code by doing python from the burdock root, and this also works for pytest (python -m pytest) from burdock root.  The Jupyter notebook can add the search path for now, since that's the only workaround I know for Jupyter before we make the module setup.

PR #85, opened 2020-01-14
Adding Mechanism Tests and Exploration Class

ankit-oss:
Added tests to pass Laplace and Gaussian mechanisms through DP predicate test. Added the tests to run via pytest.


ankit-oss:
Thank you Josh!

PR #89, opened 2020-01-15
Unit tests for empty recordsets

joshua-oss:


joshua-oss:
Print statements were there accidentally, I removed them

PR #91, opened 2020-01-16
Powerset test

### LINKED ISSUE #49: Allow full search of neighboring datasets when neighborhood is small
Evaluation should allow full search of neighboring datasets for small test datasets, where number of pairs in the powerset that differ by 1 is around 10^3.  Needs to be fast.  Could be implemented by generating all sets in the powerset as static files or tables prior to starting the validator, or by doing some trick with indices and "WHERE ... IN".

The size of the powerset is determined by the number of distinct individuals (or number of rows, if row-level privacy).

### LINKED ISSUE #92: DP predicate test failing on SUM between single record and empty database
If D1 = 100, D2 = empty database or D1 = 10000, D2 = empty database, DP predicate test is failing on SUM query. It can be seen in this [Jupyter notebook](https://github.com/privacytoolsproject/burdock/blob/powerset_test/evaluation/Differential%20Privacy%20Verification.ipynb) at the end where I run the 1_1 and 2_2 tests in Powerset test. 

1_1 means out of the 3 records {e1, e2, e3} on the top node of database search graph, D1 has the second record i.e. Usage = 100 and D2 drops that record (becoming empty)
2_2 means out of the 3 records {e1, e2, e3} on the top node of database search graph, D1 has the third record i.e. Usage = 10000 and D2 drops that record (becoming empty)

ankit-oss:
Do a scan of all neighboring pairs in database search using an initial 3 row dataset on Log scale to get the 12 edges. This fixes #49 and fixes #92 

Now powerset test is passing for all 12 edges. 

I made changes to styling (file rename, logging, deleting commented code) based on Eddie's comments in the other PR). The only thing I couldn't resolve is the sys.path.append which I couldn't get around, else it was giving me ModuleNotFound error for pytest tests. I'll look at Josh's idea of module or trying pytest from root folder.


evaluation/aggregation.py
```diff
@@ -1,64 +1,95 @@
...
+        self.mechanism = mechanism
+
+    # Taking df as a parameter it shall be passed both d1 and d2 that differ by 1 record
+    def exact_count(self, df, colname):
+        return np.zeros(self.repeat_count) + df[colname].count()
+
+    def buggy_count(self, df, colname):
+        return df[colname].count() + np.random.random_sample((self.repeat_count,))*10
+
+    def dp_count(self, df, colname):
```
joshua-oss: In a future version, we should make these be data-driven.  For example, from a manifest with a list of function handles for black box functions that are assumed to be differentially private.

evaluation/exploration.py
```diff
@@ -0,0 +1,90 @@
...
+        if(len(d1) == 0):
+            return
+        else:
+            indexes = list(d1.index.values)
+            s = [str(i) for i in indexes]
+            d1_idx_range = "".join(s)
+            for drop_idx in indexes:
+                filename = d1_idx_range + "_" + str(drop_idx)
+                if(filename not in self.visited):
+                    d2 = d1.drop(drop_idx)
```
joshua-oss: For a future implementation, you can eliminate all file access here by loading the Pandas DF directly from vectors and by creating the metadata as an object literal.  I believe PandaSQL will still use temp files for the underlying SQLite, but this may speed up the rest of the pipeline.

evaluation/aggregation.py
ankit-oss: Nice. That's a good idea!

evaluation/exploration.py
ankit-oss: I think this'll lead to substantial reduction in runtime speed. I'll work on this next. I don't really need Pandas.

PR #93, opened 2020-01-16
Eedeleon/simplify analyst scenario

eedeleon:
- CSVReader -> DataFrameReader
- add execute_private_query(reader, schema, budget, query) returns a dataframe of the sql results
- update DataFrameReader to sanitize inputs and schemas to allow for " ", "(", and ")" as well as quoted column names
- add a test for the local analyst scenario that uses execute_private_query and the DataFrameReader


sdk/burdock/query/sql/reader/__init__.py
```diff
@@ -1,3 +1,3 @@
-from .csv_reader import CSVReader
+from .dataframe_reader import DataFrameReader
 
-__al__ = ["CSVReader"]
+__al__ = ["DataFrameReader"]
```
joshua-oss: Should this be __all__?

sdk/burdock/query/sql/reader/dataframe_reader.py
```diff
@@ -0,0 +1,116 @@
...
+        for column_name in self.original_column_names:
+            sanitized_column_name = self._sanitize_column_name(column_name)
+            metadata.m_tables[table_name].m_columns[sanitized_column_name] = metadata.m_tables[table_name].m_columns[column_name]
+            metadata.m_tables[table_name].m_columns[sanitized_column_name].name = sanitized_column_name
+            has_key = has_key or metadata.m_tables[table_name].m_columns[sanitized_column_name].is_key
+            self.df[sanitized_column_name] = self.df[column_name]
+            if column_name != sanitized_column_name:
+                del metadata.m_tables[table_name].m_columns[column_name]
+
+        if not has_key:
```
joshua-oss: Nice

sdk/burdock/query/sql/reader/dataframe_reader.py
```diff
@@ -0,0 +1,116 @@
...
+
+        df_for_diffpriv1234 = self.df
+        q_result = sqldf(clean_query(query), locals())
+        return [tuple([col for col in q_result.columns])] + [val[1:] for val in q_result.itertuples()]
+
+    """
+        Executes a parsed AST and returns a typed recordset.
+        Will fix to target approprate dialect. Needs symbols.
+    """
+    def execute_typed(self, query):
```
joshua-oss: In a future update, we can rename this execute_ast_typed and add the other helper methods on base class to match query.py.

sdk/burdock/query/sql/reader/__init__.py
eedeleon: Oh wow, nice catch

PR #96, opened 2020-01-22
Switched out IBM DP lin reg with Covariance Method lin reg

mayigrin:
`lin-reg-module` now contains a differentially private implementation of linear regression that uses the covariance method.

Methods to look at would all be inside of `dp_covariance.py`. If I had to name a few I'd look at `release`, `cov_method_lin_reg`, and `amsweep`.

Also, although it doesn't say it in the git logs, commit "_Completed MVP for covariance-based diff priv lin reg_" contains work by both me and @esfahany; Kathleen is responsible for the implementation of `amsweep`, and I am responsible for the rest of the code.


service/modules/lin-reg-module/cov_method.py
```diff
@@ -0,0 +1,60 @@
+import mlflow
```
eedeleon: Leftover by mistake? I think linreg.py is the updated version

service/modules/lin-reg-module/dp_covariance.py
```diff
@@ -0,0 +1,407 @@
...
+    if accuracy <= 0:
+        raise ValueError("Privacy parameter epsilon must be a value greater than zero")
+    return accuracy
+
+
+def laplace_get_accuracy(sens, epsilon, alpha=.05):
+    return np.log(1 / alpha) * (sens / epsilon)
+
+
+def distribute_epsilon(globalEps, nCalcs=None, epsilonDist=None):
```
eedeleon: minor comment but python convention for parameters is snake_case.

service/modules/lin-reg-module/dp_covariance.py
```diff
@@ -0,0 +1,407 @@
+import pandas as pd
+import numpy as np
+import math
+
```
eedeleon: Reviewing this file will require some prep on my side. I will get to it tomorrow

service/modules/lin-reg-module/dp_lin_reg.py
```diff
@@ -0,0 +1,91 @@
...
+    3.0000...
+    >>> reg.predict(np.array([[3, 5]]))
+    array([16.])
+    """
+
+    def __init__(self, fit_intercept=False):
+        self.fit_intercept = fit_intercept
+
+    def _set_coef_and_intercept(self, df):
+        df_copy = df
```
eedeleon: this won't protect the original df from having a column dropped from what I can tell. Do you need a deepcopy?

service/modules/lin-reg-module/dp_lin_reg.py
```diff
@@ -0,0 +1,91 @@
...
+        if self.fit_intercept:
+            self.intercept_ = df_copy['intercept', 'Estimate']
+            df_copy = df_copy.drop(['intercept'])
+        self.coef_ = df_copy['Estimate'].values
+
+    def fit(self, X, y, rng, budget):
```
eedeleon: we usually call rng bounds, range is a keyword so it is helpful

eedeleon: I suspect rng was inherited from the R implementation, I think bounds for python works better since it avoids the keyword

service/modules/lin-reg-module/dp_lin_reg.py
mayigrin: image

I had the same worries so I actually tested that before I put the code in? And it looks like it does protect the original. But if you'd like I can still make it a deepcopy regardless.

eedeleon: nice check

eedeleon:
Great work both of you. I will merge this so we have a simpler time demoing it. We will need to find a valid way of testing it to catch bugs but otherwise it is very clean code and the comments match the code well.

PR #99, opened 2020-01-27
Add python version of _plain_sgd and co

eedeleon:
- Check for custom_mlflow_setup.py on startup, if available, call mlflow_setup() from it. Allows service deployers to create a remote mlflow tracking server and save releases


eedeleon:
More progress was made on https://github.com/privacytoolsproject/burdock/pull/101

Issue #97, opened 2020-01-23
POWER function encountering ValueError exception


ankit-oss:
**Repro Steps**

import evaluation.aggregation as agg
dv = DPVerification(dataset_size=500)
ag = agg.Aggregation(t=1, repeat_count=10000)
d1_query = "SELECT POWER(SUM(Usage), 2) AS UsageSquared FROM d1.d1"
d2_query = "SELECT POWER(SUM(Usage), 2) AS UsageSquared FROM d2.d2"
dp_res = dv.dp_groupby_query_test(d1_query, d2_query, plot=True, repeat_count=100)

Exception happens at this line - 
newrs = TypedRowset([colnames], types, sens)

It is because types is of type 'unknown'
types = [s[1].type() for s in syms]

**Error**
Exception has occurred: ValueError
Trying to load unknown type unknown
  File "/e/Git/burdock/sdk/burdock/query/sql/reader/rowset.py", line 105, in __setitem__
    raise ValueError("Trying to load unknown type " + t)
  File "/e/Git/burdock/sdk/burdock/query/sql/reader/rowset.py", line 55, in __init__
    self[cn] = tempcols[cn]
  File "/e/Git/burdock/sdk/burdock/query/sql/private/query.py", line 120, in _postprocess
    newrs = TypedRowset([colnames], types, sens)
  File "/e/Git/PrivacyToolsProject/burdock/evaluation/aggregation.py", line 104, in run_agg_query_df
    sample_res = private_reader._postprocess(subquery, query, syms, types, sens, srs)
  File "/e/Git/PrivacyToolsProject/burdock/evaluation/dp_verification.py", line 315, in dp_groupby_query_test
    d1_res, dim_cols, num_cols = ag.run_agg_query_df(d1, d1_metadata, d1_query, confidence, file_name = "d1")
  File "/e/Git/PrivacyToolsProject/burdock/evaluation/dp_verification.py", line 383, in main
    dp_res = dv.dp_groupby_query_test(d1_query, d2_query, plot=True, repeat_count=100)
  File "/e/Git/PrivacyToolsProject/burdock/evaluation/dp_verification.py", line 402, in <module>
    print(dv.main())

PR #103, opened 2020-02-11
Pull through symbols for postprocess functions

joshua-oss:
Bug fix for #97.  Some post process functions were passing through without connecting to the source columns.  Fixed bug and added two new exceptions to catch any case where a function forgets to implement the symbol() call.  Added unit test.

### DIFF
--- a/sdk/burdock/query/sql/ast/expressions/numeric.py
+++ b/sdk/burdock/query/sql/ast/expressions/numeric.py
@@ -48,8 +48,6 @@ class ArithmeticExpression(SqlExpr):
         self.left = left
         self.right = right
         self.op = op
-    def symbol(self, relations):
-        return ArithmeticExpression(self.left.symbol(relations), self.op, self.right.symbol(relations))
     def type(self):
         if self.op == "/":
             return "float"
@@ -93,6 +91,8 @@ class ArithmeticExpression(SqlExpr):
         l = self.left.evaluate(bindings)
         r = self.right.evaluate(bindings)
         return ops[self.op](l, r)
+    def symbol(self, relations):
+        return ArithmeticExpression(self.left.symbol(relations), self.op, self.right.symbol(relations))
 
 class MathFunction(SqlExpr):
     def __init__(self, name, expression):
@@ -118,6 +118,8 @@ class MathFunction(SqlExpr):
     def evaluate(self, bindings):
         exp = self.expression.evaluate(bindings)
         return funcs[self.name.lower()](exp)
+    def symbol(self, relations):
+        return MathFunction(self.name, self.expression.symbol(relations))
 
 class PowerFunction(SqlExpr):
     def __init__(self, expression, power):
@@ -130,6 +132,8 @@ class PowerFunction(SqlExpr):
     def evaluate(self, bindings):
         exp = self.expression.evaluate(bindings)
         return np.power(exp, self.power.value)
+    def symbol(self, relations):
+        return PowerFunction(self.expression.symbol(relations), self.power.symbol(relations))
 
 class BareFunction(SqlExpr):
     def __init__(self, name):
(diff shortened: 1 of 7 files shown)


sdk/burdock/query/sql/ast/expressions/numeric.py
```diff
@@ -48,8 +48,6 @@ def __init__(self, left, op, right):
         self.left = left
         self.right = right
         self.op = op
-    def symbol(self, relations):
```
joshua-oss: Just moving this function down in the class definition

sdk/burdock/query/sql/ast/expressions/sql.py
```diff
@@ -4,37 +4,6 @@
     SQL-specific expressions
 """
 
-class Column(SqlExpr):
```
joshua-oss: Move this class to Tokens

sdk/burdock/query/sql/ast/tokens.py
```diff
@@ -188,6 +188,12 @@ def evaluate(self, bindings):
         raise ValueError("We don't know how to evaluate " + str(self))
     def children(self):
         return [None]
+    def symbol(self, relations):
```
eedeleon: Where is this new method used?

joshua-oss: This method was always part of the code, but I forgot to override it on some of the AST nodes.  The behavior of symbol() and load_symbols() is specified in the dev spec at https://drive.google.com/file/d/1XHZUiYpYcMP6Ix56lq-Og6iU33GI8cqV/view.  The "AST" heading gets into load_symbols, which is what resolves curator-supplied symbols, and the "Expressions" heading talks about symbol() which is what resolves analyst-provided symbols.  Those two methods basically connect analyst and curator symbols, since both sources are externally-provided and can have various inconsistencies.

The bug here was that I implemented, symbol() on the base classes to just pass through.  That was great for AST nodes that don't actually reference any database columns.  However, some AST nodes reference database columns, maybe nested way down the tree, and passing them through without recursively walking was leaving the database column symbols unresolved.

On the lowest base class, I now only pass through if the node doesn't reference database columns.  If it references database columns, it means we should've overriden the method on the specialized class.  So I throw in that case.  I also added this case on bare "Expression", because the bare expression is a pass-through that we can't even check to see if it has children.  We should almost never actually use this class anyway; it's like an escape valve for when we haven't implemented a more specialized expression in the parser.  Since we can't verify if it's safe to pass through the symbol, we throw here.  Both exceptions should never be hit.

I also went through and added the symbol everywhere I forgot it.  It's worth doing another pass to make sure we didn't miss any.  Anywhere that might reference a database column needs to have a .symbol() attached.  For example, people might use a database column to set the exponent on a power, even though it's a literal 99.9% of the time, we still need to try to resolve it.

This also raises another point about about how we do these type of operations in the code.  There are at least 3 places like this where we are traversing the tree and doing things based on the values.  My current pattern in all these cases is to use recursion.  That makes it important to put in safeguards like the exceptions I added here, in case we forget to implement the recursion methods on an inherited class.

In Scala, I would probably instead do a deep copy with pattern matching to swap out the changed nodes as we copy.  We could hack together something like that for Python, but it doesn't feel right.  Maybe there are benefits to that approach though.

Another way to do it would be like what we could do in C# or Java, where we locate the target nodes in the tree using an xpath-style accessor, (e.g. .findnodes(Column)) and then use the references to modify them in-place.  I implemented the two tree-accessor functions and use them in multiple places, but Python doesn't seem to like in-place editing of objects returned that way.  Or maybe I just don't understand how to return references rather than copies?  Anyway, this pattern has some advantages and disadvantages, but we could maybe make that work.

I'm OK with sticking with recursion as well; just open to rethinking it.

ankit-oss:
Ran the DP powerset test on POWER(SUM(Usage), 2) and it passes now.

PR #104, opened 2020-02-11
Eedeleon/add dataverse loader

eedeleon:
Add dataverse loader as a utility to read data from dataverse into a dataframe


eedeleon:
- add a dataverse loader utility function for users under burdock.data
- expose metadata creation classes under burdock.metadata

PR #114, opened 2020-02-19
Eedeleon/add lin reg to burdock

eedeleon:


eedeleon:
Add demo notebooks and lin regression code to burdock. No linear regression demo is added because all runs of DPLinReg are failing with "Matrix is not positive definite"

PR #118, opened 2020-02-25
Data readers

joshua-oss:
Adds data readers for Spark, Postgres, SqlServer, Presto.  Renames DataFrameReader to PandasReader.

Updates name compare attach on Collection object to be dynamic, based on engine property specified in metadata.  If engine not specified, base name compare is used


sdk/burdock/metadata/collection.py
```diff
@@ -7,12 +8,18 @@ def __init__(self, tables, engine, compare=None):
...
         if compare is None:
-            if engine == "Postgres":
-                self.compare = PostgresNameCompare()
-            elif engine == "MSSQL":
-                self.compare = MSSQLNameCompare()
-            else:
+            if engine == "Unknown":
                 self.compare = BaseNameCompare()
+            else:
+                engine_module_name = "burdock.query.sql.reader." + engine.lower() + "_reader"
```
eedeleon: What is the goal of this code?

sdk/burdock/query/sql/reader/spark_reader.py
```diff
@@ -0,0 +1,74 @@
...
+from burdock.metadata.name_compare import BaseNameCompare
+from .rowset import TypedRowset
+
+from burdock.query.sql.ast.tokens import Literal
+from burdock.query.sql.ast.expressions.numeric import BareFunction
+
+class SparkReader:
+    def __init__(self, host, database, user, password=None, port=None):
+        from pyspark.sql import SparkSession
+        self.api = SparkSession.builder.config("spark.rpc.askTimeout", "600s").config("spark.executor.memory", "6g").config("spark.driver.memory", "6g").config("spark.executor.memoryOverhead", "600").config("spark.driver.memoryOverhead", "600").getOrCreate()
```
eedeleon: We can check this PR as is but will need to update it for configuration support

joshua-oss: Oh, shoot, that's a bug.  I wanted to move that out and have the caller pass in the context (like Dylan's sample in email earlier today).  Will fix and update

sdk/burdock/metadata/collection.py
joshua-oss: The metadata has an "engine" property that specifies which database engine is being used, and that determines how analyst-provided names are matched to database names.  For example, in Postgres, names can be case-sensitive, and in SqlServer, names with spaces can be escaped with square brackets.

The previous version of the code was hard-coded to just support name schemes for Postgres and SqlServer, with a if/elif switch block.  This code allows any reader that is installed in the SDK to be used for engine-specific name comparison semantics, as long as it follows the naming conventions.  So, if the metadata specifies that the tables use Postgres naming conventions, it will automatically uses PostgresNameCompare() from postgres_reader.

The spec for name-matching semantics is at https://drive.google.com/file/d/1XHZUiYpYcMP6Ix56lq-Og6iU33GI8cqV/view, and addition of the "engine" property was done so that we could avoid having to make the system transparently map between various different incompatible semantics.

sdk/burdock/query/sql/reader/spark_reader.py
eedeleon: passing in session solved this

sdk/burdock/metadata/collection.py
eedeleon: Makes sense. traditionally I would have this done with a registration step.

class MyReader(Reader):

Reader.register_engine(MyReader)

Then for the func:
Reader.get_func(engine) # returns the name compare func

eedeleon: class Reader:
      _name_compare_funcs = {}
    @classmethod
     def register_name_compare(cls, class_to_add):
         cls._name_compare_funcs[class_to_add.name] = class_to_add.name_compare
     def get_name_compare(cls, key): return cls._name_compare_funcs[key]


try:
      self.compare = Reader.get_name_compare(engine)
except KeyError :
      self.compare = BaseNameCompare()

eedeleon: I usually wrap it in a func instead of keying the dictionary because it allows for a lock if one is needed

joshua-oss: I like this pattern.  If I'm understanding correctly, you're using the base class as a static dictionary for the compare functions on all derived classes.

In this case, the curator-supplied names and the analyst-supplied names are intended to be able to be matched without needing a Reader, though.  The thinking was that the analyst, curator, and data processor are three different entities, and the data processor is the only entity that needs a reader.  The analyst and curator just need a way to ensure that names can be resolved.  So, for example, if the curator names a table something like [SalesTotals] in the metadata, and the analyst specified a query like ".... WHERE SALESTOTALS.key = otherTable.key", the validator and everything should be able to figure out they're the same without needing any sort of Reader.  The Reader is used to hold transient connection string information and presumably lives in the hardened middleware.  And the analyst isn't supposed to need to know what the engine is; since that's specified by the data curator.  So the goal is for the analyst-provided SQL strings to be able to validate against the curator-supplied metadata, do recursive type inference and in-memory expression evaluation, etc. without a database connection.

And since the engine semantics are determined by the curator, the analyst doesn't need to instantiate a name comparer or reader or anything, the idea is that the AST automatically uses the right semantics as specified in the metadata file.

eedeleon: I am not sure if the curator will always specify the engine semantics. Either way, I think the static dictionary + engine string key into it would still allow us to make readers an internal detail. 

If the key comes from the metadata, it would work the same way as the current implementation. With the benefit on not enforcing a naming convention on derived classes from instantiating the base class.

PR #127, opened 2020-02-25
Cleaning old evaluator tests and disabling Yarrow

ankit-oss:
Disabled Yarrow for now till it starts working
Removing some aggregation tests from main
Fixing group by test result


ankit-oss:
This is incorporated in the other PR #133

PR #131, opened 2020-02-26
Naming updates

joshua-oss:
Move and rename for usability:

burdock.query.sql -> burdock.sql
burdock.query.sql.reader -> burdock.reader.sql

PrivateQuery -> PrivateReader
Collection -> CollectionMetadata
MetadataLoader -> CollectionMetadata.from_file()
BaseNameCompare -> NameCompare

All readers now inherit three utility methods.  Implementer implements execute() which takes string and returns untyped rowset of tuples for compat with DBAPI.  Other three methods are:
* execute_typed() - takes string and returns typed rowset, infer types from return tuple
* execute_ast() - takes compiled AST and executes, returning untyped tuples
* execute_ast_typed() - executes typed AST and returns typed rowset, with types pulled from resolved symbols on AST


evaluation/dp_verification.py
```diff
@@ -83,7 +83,7 @@ def generate_neighbors(self, load_csv = False):
         d1_table.schema, d2_table.schema = "d1", "d2"
         d1_table.name, d2_table.name = "d1", "d2"
         d2_table.rowcount = d1_table.rowcount - 1
-        d1_metadata, d2_metadata = Collection([d1_table], "csv"), Collection([d2_table], "csv")
```
eedeleon: the import change?

samples/DP_SQL_dataverse_dataset.ipynb
```diff
@@ -24,8 +24,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from burdock.query.sql import PandasReader, execute_private_query\n",
-    "from burdock.metadata.collection import Collection, Table, Float"
+    "from burdock.sql import PandasReader, execute_private_query\n",
+    "from burdock.metadata.collection import CollectionMetadata, Table, Float"
```
eedeleon: burdock.metadata import Colle*** ?

sdk/burdock/data/adapters.py
```diff
@@ -65,7 +65,7 @@ def load_metadata(dataset_document):
 
 def load_reader(dataset_document):
     """
-    rtype: BaseReader
+    rtype: Base
```
eedeleon: Reader?

samples/DP_SQL_dataverse_dataset.ipynb
joshua-oss: CollectionMetadata is also in __init__ for burdock.sql.  I don't think we should add all of the other ones, since there would be name collisions, so assumption is that we import burdock.metadata.collection * only when doing literals.

sdk/burdock/data/adapters.py
joshua-oss: Doh!

evaluation/dp_verification.py
joshua-oss: Yes, that code works; the import is line 20

PR #134, opened 2020-02-28
Add databrickSparkReader to enable larger data set DP execution

dylanw-oss:
I met a circular namespace bug and can't test the code.  But I did tested it with the code without several PRs. And it works.  (update, resolve it in 3rd commit)

To make it work, I need some helps on subquery and query.

### DIFF
--- a/sdk/burdock/reader/sql/databricks.py
+++ b/sdk/burdock/reader/sql/databricks.py
@@ -1,0 +1,97 @@
+import datetime
+from pyspark.sql import Row
+from pyspark.sql.types import BooleanType
+from burdock.sql import CollectionMetadata
+from burdock.sql import PrivateReader
+from burdock.mechanisms.gaussian import Gaussian
+
+class DatabricksSparkReader:
+    def __init__(self, sparkSession, meta_path):
+        self.spark = sparkSession
+        self.schema = CollectionMetadata.from_file(meta_path)
+        self.private_reader = PrivateReader(None, self.schema, 1.0)
+
+    """
+        Executes a raw SQL string against the Spark Hive database and returns spark dataframe.
+    """
+    def execute(self, query):
+        if not isinstance(query, str):
+            raise ValueError("Please pass strings to execute.  To execute ASTs, use execute_typed.")        
+        
+        max_contrib = self.private_reader.max_contrib
+        epsilon = self.private_reader.epsilon
+        interval_widths = self.private_reader.interval_widths
+
+        subquery, fullquery = self.private_reader.rewrite(query)
+        syms = subquery.all_symbols()
+        numeric_cols = subquery.numeric_symbols()
+
+        if numeric_cols is None:
+          pass
+
+        subquery_df = self.spark.sql(str(subquery))
+        noisedDF = subquery_df.rdd.mapPartitions(applyNoises(numeric_cols, epsilon, max_contrib, interval_widths)).toDF()
+        
+        # censor dimensions for privacy
+        if subquery.agg is not None:
+            noisedDF = noisedDF.filter(noisedDF.keycount > max_contrib ** 2)
+
+        noisedDF.createOrReplaceTempView("query_df_noised")
+        noisedDF.cache()          
+
+        q_result = self.spark.sql(str(fullquery))
+        return q_result
+
+"""
+    Mappartition function as a trasnformer only accepts a function with iterator as the single parameter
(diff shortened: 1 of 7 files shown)


sdk/burdock/reader/sql/databricks.py
```diff
@@ -0,0 +1,97 @@
...
+        interval_widths = self.private_reader.interval_widths
+
+        subquery, fullquery = self.private_reader.rewrite(query)
+        syms = subquery.all_symbols()
+        numeric_cols = subquery.numeric_symbols()
+
+        if numeric_cols is None:
+          pass
+
+        subquery_df = self.spark.sql(str(subquery))
```
dylanw-oss: (1) subquery
I need some help to change subquery for databricks spark.
a. remove "per_key_randome.row_num <= 5", I think it's a bug as we observed, if users' data have 6 months of data, and groupby date, will be restricted to return at most 5 months of data.
b. from "Random" to "Rand"

sdk/burdock/reader/sql/databricks.py
```diff
@@ -0,0 +1,97 @@
...
+        
+        # censor dimensions for privacy
+        if subquery.agg is not None:
+            noisedDF = noisedDF.filter(noisedDF.keycount > max_contrib ** 2)
+
+        noisedDF.createOrReplaceTempView("query_df_noised")
+        noisedDF.cache()          
+
+        q_result = self.spark.sql(str(fullquery))
+        return q_result
```
dylanw-oss: (2) fullquery
I need some help to change fullquery for databricks spark.
basically replace "subquery" with a temp table string.
for example, for PUMs example, it can be 
"SELECT ( SUM ( count_0xe3f0 ) ) AS c, married AS m FROM **query_df_noised** AS exact_aggregates GROUP BY married ORDER BY c, m DESC"
replace the subquery with "query_df_noised"

sdk/burdock/reader/sql/databricks.py
```diff
@@ -0,0 +1,97 @@
+import datetime
+from pyspark.sql import Row
```
eedeleon: We might be able to remove this dependency if we can use map reduce? udf might be a problem. Requiring spark for all scenarios won't be possible but we can require spark for this addition as a temporary solution. It may be worth having a burdock-spark package longer term since it looks like the spark ecosystem is perfect for DP.

sdk/burdock/reader/sql/databricks.py
```diff
@@ -0,0 +1,97 @@
+import datetime
```
eedeleon: Awesome PR.  Just to get an idea, is this databricks specific because of some nuances in the implementation or could this work on other clusters? I am trying to see if this should be named SparkReader or not.

sdk/burdock/reader/sql/databricks.py
eedeleon: @joshua-oss  is working on making tau thresholding optional which should help with 1.

We can help guide you through adding a custom serializer that resolves the Random -> Rand issue

sdk/burdock/reader/sql/databricks.py
dylanw-oss: Thanks @eedeleon you're right, it makes sense. 
I named it as databrickSparkReader just because SparkReader already added by @joshua-oss :), it is the old one we developed previously and it only works with a dataset size fits in memory, not as good as the new one we add now. But it matches everything we have in new semantics right now, 
and I can't pass dummy None to PrivateReader, I have to use dummy SparkReader, I made a new commit to fix it, basically use SparkReader in databrickSparkReader. 
Probably we will have both in parallel for now.

Do you have other suggestion on the name?

sdk/burdock/reader/sql/databricks.py
dylanw-oss: Agreed! Probably I can change it to return iterator in mapPartition function and remove this dependency. I'll try it.
As the first version, and it's working well, we can leave it as is for now.

sdk/burdock/reader/sql/databricks.py
dylanw-oss: @joshua-oss had the PR sent, will take a look, I need incorporate his change for sure, in next iteration PR :)

dylanw-oss:
It will be moved to https://github.com/privacytoolsproject/burdock-samples

PR #135, opened 2020-02-28
Private Reader Options

joshua-oss:
Implement 4 options:

* reservoir_sample
* clamp_counts
* censor_dimensions
* clamp_columns

Documentation at:
https://github.com/privacytoolsproject/burdock-samples/blob/master/docs/data_access/options.md


sdk/burdock/sql/private_rewriter.py
```diff
@@ -148,11 +150,16 @@ def exact_aggregates(self, query):
...
         subquery = Query(child_scope.select(), query.source, query.where, query.agg, query.having, None)
-        subquery = self.per_key_random(subquery)
-        subquery = [AliasedRelation(subquery, "per_key_random")]
+        if self.options.reservoir_sample:
+            subquery = self.per_key_random(subquery)
+            subquery = [AliasedRelation(subquery, "per_key_random")]
 
-        filtered = Where(BooleanCompare(Column("per_key_random.row_num"), "<=", Literal(str(self.max_contrib), self.max_contrib)))
-        return Query(select, From(subquery), filtered, query.agg, None, None)
+            filtered = Where(BooleanCompare(Column("per_key_random.row_num"), "<=", Literal(str(self.options.max_contrib), self.options.max_contrib)))
```
dylanw-oss: Can we disable this filter in subquery through options?

sdk/burdock/sql/private_rewriter.py
```diff
@@ -148,11 +150,16 @@ def exact_aggregates(self, query):
...
+            return Query(select, From(subquery), filtered, query.agg, None, None)
+        else:
+            subquery = self.per_key_clamped(subquery)
+            subquery = [AliasedRelation(subquery, "per_key_all")]
+            return Query(select, From(subquery), None, query.agg, None, None)
```
dylanw-oss: For databricks spark, how to return 

"SELECT .... FROM (subquery_table) ..."

instead of a subquery string, just a string representing the return data set, as in databricksSparkReader, I cache the subquery result in a temp view table.

I think you're working on this?

sdk/burdock/sql/private_reader.py
```diff
@@ -224,3 +224,17 @@ def execute_ast_typed(self, query):
...
+
+class PrivateReaderOptions:
+    def __init__(self, 
+        censor_dims=True, 
+        clamp_counts=True, 
+        reservoir_sample=True,
+        clamp_columns=True,
+        row_privacy=False):
+
+        self.censor_dims = censor_dims
```
dylanw-oss: param description?

I noticed some params shared the same in RewriterOptions, is it possible to combine them together?

sdk/burdock/sql/private_rewriter.py
joshua-oss: Yes, that will automatically be skipped if you do private.options.reservoir_sample = False

sdk/burdock/sql/private_rewriter.py
joshua-oss: Yep, that will be in a future PR

sdk/burdock/sql/private_reader.py
joshua-oss: I updated the docstrings.

PR #137, opened 2020-02-28
Bias test

ankit-oss:
Adding a test which computes mean signed deviation and outputs it as a ratio of actual value. If ratio is outside +/-5% of actual value (relaxed compared to 1% where it does fail quite a bit), we raise a bug as any result with so much bias on repeated querying is quite hard to consume. 

Bias becomes quite significant as records go lower, for example in Powerset test which fails right now. This is quite expect as we do want datasets with such low size to not extract meaningful unbiased results. 

Will add a hypothesis test with p-value in next PR for SUM, COUNT, MEAN, VAR operators.


ankit-oss:
Thank you!

PR #145, opened 2020-03-07
Map-based processing

joshua-oss:


joshua-oss:
* Same code path supports map, sort, and filter against Python iterators in-memory or distributed via RDD or DF
* pre/post is replaced with internal method that can cache exact and call repeatedly (unit test has example)
* Fixed bug where GROUP BY on numeric column was adding noise
* Removed sensitivity and sort from TypedRowset, they're not needed.

TODO:
- Removed confidence intervals, will add back as output columns like colname_095_low
- internal convert() function isn't necessarily robust yet.  Would be useful for cracking CSVs, stripping quote characters, doing date conversions, etc.  Should make it a library function that can be called in a map(zip())

joshua-oss:
Some updates:

* row_privacy works now, so tables with no keys can be used
* non-key counts can be clamped to zero
* COUNT DISTINCT works properly again
* reservoir sampling is for the whole query
* analyst can pass in a max_contrib that is more aggressive than curator-supplied

PR #150, opened 2020-03-10
Override execute_typed in SparkReader

dylanw-oss:


dylanw-oss:
@joshua-oss, I added the override in SparkReader.

PR #152, opened 2020-03-10
Eedeleon/expose report top level

eedeleon:
- In a push for top level directories to have tree like dependencies we are moving report up a level

Current hierarchy:
- sql(ast, report, reader, metadata)
- reader(report)
- metadata(reader)
- mechanisms(report)
- eval(sql, reader)
- models(report)
- client
- ast
- report

Previous hierarchy:
- sql(ast, reader, metadata)
- reader(metadata) 
- metadata(reader)
- eval(sql, reader)
- mechanisms
- models
- client
- ast


dylanw-oss:
@eedeleon , I think there're some other places you missed.

search "burdock.metadata.report", there're 6 places need update, for example,
gaussin.py
...
from burdock.metadata.report import Result, Interval, Intervals

PR #156, opened 2020-03-11
move sql readers to sql.reader, add hook in burdock.sql. 

eedeleon:
New import style:
from burdock.sql  import PrivateReader, PandasReader


tests/query/test_interval_expr.py
```diff
@@ -23,7 +22,7 @@
 class TestQuery:
     def test_group_by_noisy_typed_order_inter(self):
         reader = PandasReader(schema, df)
-        private_reader = PrivateReader(reader, schema, 1.0)
+        private_reader = PrivateReader(schema, reader, 1.0)
```
dylanw-oss: IMO, it is better the reader as the first parameter, PrivateReader felt like a decorator, and maybe schema can be designed as an optional parameter later?

joshua-oss: We wanted to be consistent with how PandasReader does it, so the metadata and the thing the reader is wrapping always come in the same order.  The other issue is that metadata can't really be optional on PrivateReader, since most readers we are wrapping don't have it.  For example, a SqlServerReader or SparkReader don't know anything about differential privacy, and they just return DP-API compatible lists of tuples.  The PrivateReader, on the other hand, needs to know metadata about private_id and sensitivity of columns.

The one case where we don't need metadata is if we are doing row_privacy and only supporting counting queries.  But I don't think we want to make that an easy onboarding path, because people would rapidly hit a brick wall

dylanw-oss: makes sense. 
Is there an easy way to create metadata rather than manually doing it?

joshua-oss: We have a code sample in the old repo that walks a database schema and infers metadata, and we could do the same for a pandas dataframe or spark dataframe that has types.  We were worried about people not understanding that min/max will leak privacy, so they need domain knowledge to set the sensitivity.  But I agree it's a pain for new users to deal with, and we should at minimum help out with some code samples soon

eedeleon: There is an example here: https://github.com/opendifferentialprivacy/whitenoise-system/blob/master/tests/test_local_analyst.py

PR #165, opened 2020-03-18
policy laplace dpsu

anujdesai510:
python implementation and unit tests for Policy Laplace DPSU: https://arxiv.org/pdf/2002.09745.pdf


sdk/opendp/whitenoise/data/dpsu.py
```diff
@@ -0,0 +1,117 @@
...
+import random
+import operator
+from collections import defaultdict
+
+from opendp.whitenoise.metadata import CollectionMetadata
+from opendp.whitenoise.sql import PandasReader
+from opendp.whitenoise.sql.parse import QueryParser
+from opendp.whitenoise.mechanisms.rand import laplace
+
+def preprocess_df_from_query(meta_path, csv_path, query, table_name):
```
eedeleon: inputs should be schema, pandas dataframe

It is the same but a bit more flexible since it won't require users to write their in memory objects to disk. It is also a one liner(as shown here) to read it from disk so it isn't too cumbersome

tests/sdk/data/test_dpsu.py
```diff
@@ -0,0 +1,34 @@
...
+        query = "SELECT married, educ FROM PUMS.PUMS GROUP BY married, educ"
+        df = preprocess_df_from_query(meta_path, csv_path, query, "PUMS.PUMS")
+
+        assert df is not None
+        assert(len(df["group_cols"][0]) == 2)
+
+    def test_dpsu_df(self):
+        original_df = pd.read_csv(csv_path)
+        query = "SELECT married, educ FROM PUMS.PUMS GROUP BY married, educ"
+        df = preprocess_df_from_query(meta_path, csv_path, query, "PUMS.PUMS")
```
eedeleon: What is our intended analyst experience?
Would it be something like:
dpsu_df = run_dpsu(df, metadata, query, eps, max_contrib, delta=some constant/based off size of df)

Can max contrib be defaulted @joshua-oss ?

sdk/opendp/whitenoise/data/dpsu.py
```diff
@@ -0,0 +1,117 @@
...
+	preprocessed_df = pd.DataFrame()
+	preprocessed_df[kc] = df[kc]
+	preprocessed_df["group_cols"] = tuple(df[group_cols].values.tolist())
+
+	return preprocessed_df
+
+def reservoir_sample(iterable, max_contrib):
+	reservoir = []
+	for i, item in enumerate(iterable):
+		if i < max_contrib:
```
eedeleon: python usually tabs in by 4 spaces not 8

tests/sdk/data/test_dpsu.py
anujdesai510: per meeting, default max_contrib to 5, default delta and see if you can get table name from metadata

sdk/opendp/whitenoise/data/dpsu.py
lurosenb: +1

sdk/opendp/whitenoise/data/dpsu.py
```diff
@@ -0,0 +1,117 @@
...
+	return df
+
+def dpsu_df(df1, df2):
+	"""
+
+	"""
+	dpsu_df = pd.merge(df1, df2, on=df2.columns[0])
+	dpsu_df.drop(["group_cols", "hash"], axis=1, inplace=True)
+
+	return dpsu_df
```
lurosenb: PR looks awesome Anuj - nice work : )

sdk/opendp/whitenoise/data/dpsu.py
```diff
@@ -0,0 +1,117 @@
...
+	group_cols = [ge.expression.name for ge in query_ast.agg.groupingExpressions]
+	kc = schema[table_name].key_cols()[0].name
+
+	preprocessed_df = pd.DataFrame()
+	preprocessed_df[kc] = df[kc]
+	preprocessed_df["group_cols"] = tuple(df[group_cols].values.tolist())
+
+	return preprocessed_df
+
+def reservoir_sample(iterable, max_contrib):
```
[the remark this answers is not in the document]
joshua-oss: The sample really just needs to be a uniform sample without replacement.  The reservoir sampling approach is a way to obtain a uniform sample without replacement when the number of elements to choose from is unknown ahead of time and unlikely to fit in memory.  Since you have all of the elements in the grouping anyway, you can just randomly do something like np.choose.

I forgot to mention the source of randomness should be cryptographic, though, so np.choose won't be random enough.  Another way to get a uniform random sample without replacement is to shuffle the set and grab the first 'k' elements.  But again it is important in that case to shuffle in a way that is cryptographically random rather than using a PRNG.

I believe that is very low risk in this case, FWIW, since I think the shuffling by user, which is accomplished up front via the hash, is more important for this algorithm.  But we should do all of them with strong randomness

peteroupc:
I should note that this mechanism (as with the current implementation of `mechanisms.rand`) relies on random floating-point numbers.  As Mironov (2012, "On Significance of the Least Significant Bits For Differential Privacy"), there can be security risks involved with random floating-point numbers that are not present with random fixed-point numbers or integers, and Mironov demonstrates this in the case of Laplacian noise.

joshua-oss:
> I should note that this mechanism (as with the current implementation of `mechanisms.rand`) relies on random floating-point numbers. As Mironov (2012, "On Significance of the Least Significant Bits For Differential Privacy") shows in the case of Laplacian noise, there can be security risks involved with random floating-point numbers that are not present with random fixed-point numbers or integers.

Good point.  The intent is to replace these mechanisms with the implementations from core, and I believe Christian has implemented the clamping mechanism for that Laplace implementation.  Python bindings are available again as of this week, so we will make a GitHub issue to make sure we take care of this.

sdk/opendp/whitenoise/sql/dpsu.py
```diff
@@ -0,0 +1,117 @@
...
+    table_name = query_ast.m_sym_dict[group_cols[0]].tablename
+    key_col = schema[table_name].key_cols()[0].name
+
+    preprocessed_df = pd.DataFrame()
+    preprocessed_df[key_col] = df[key_col]
+    preprocessed_df["group_cols"] = tuple(df[group_cols].values.tolist())
+
+    return preprocessed_df
+
+def reservoir_sample(iterable, max_contrib):
```
eedeleon: Not sure what iterable means here. Can you explain it?

sdk/opendp/whitenoise/sql/dpsu.py
```diff
@@ -0,0 +1,117 @@
...
+
+        cost_dict = {}
+        for item in items:
+            if histogram[item] < gamma:
+                cost_dict[item] = gamma - histogram[item]
+
+        budget = 1
+        k = len(cost_dict)
+
+        cost_dict = {k: v for k, v in sorted(cost_dict.items(), key=lambda item: item[1])}
```
eedeleon: sorted_items = [k for k, v in sorted(cost_dict.items(), key=lambda item: item[1])]

eedeleon: no need to rewrite cost_dict

sdk/opendp/whitenoise/sql/dpsu.py
```diff
@@ -0,0 +1,117 @@
...
+
+def run_dpsu(schema, input_df, query, eps, delta=math.exp(-10), max_contrib=5):
+    preprocessed_df = preprocess_df_from_query(schema, input_df, query)
+    
+    dpsu_df = policy_laplace(preprocessed_df, eps, delta, max_contrib)
+
+    output_df = pd.merge(input_df, dpsu_df, on=dpsu_df.columns[0])
+    output_df.drop(["group_cols", "hash"], axis=1, inplace=True)
+
+    return dpsu_df
```
eedeleon: return output_df right?

PR #167, opened 2020-03-20
Feat release dataset

lurosenb:
Ready for merge : )
- Added ReleaseDatasetDocument + DatasetReadReleaseRequest
- Added guid headers for client
- Added client support for read_released() and release()
- Added notion of authorized users (reference implementation)
- Added example released datasets
- Added USERS_RELEASED and OWNERS (reference implementations)
- Added dataset.py implementations for release and readreleased
- Updated register/read implementations to use helpers
- Added register/release module
- Added new tests including one system test


openapi/swagger.yml
```diff
@@ -109,6 +147,16 @@ definitions:
         type: number
         format: float
         description: Budget used by this read request
+  DatasetReleaseRequest:
```
eedeleon: This assumes the original dataset was private? Should it?

openapi/swagger.yml
```diff
@@ -109,6 +147,16 @@ definitions:
...
+    type: object
+    properties:
+      dataset_name:
+        type: string
+        description: The name for the dataset
+      budget:
+        type: number
+        format: float
+        description: Override budget for the released dataset
```
eedeleon: Set budget for the released dataset

sdk/opendp/whitenoise/client/__init__.py
```diff
@@ -34,6 +35,17 @@ def submit(self, params, uri):
...
 
 class DatasetClient(object):
+    def readreleased(self, dataset_name, budget):
+        client = _get_client()
+        read_released_request = DatasetReadRequest(dataset_name=dataset_name, budget=budget)
+        return client.datasetreadreleased(read_released_request)
+
+    def release(self, dataset_name, budget=None):
+        client = _get_client()
+        release_request = DatasetReleaseRequest(dataset_name=dataset_name, \
```
eedeleon: keeping this on one line is fine. We can add a tox.ini at the top level directory to specify the flake8 settings. We can use 100 characters as our max line length for now, this will for starters hopefully avoid your editor auto adding \. But over time it will allow us to reintroduce the flake8 build

eedeleon: https://flake8.pycqa.org/en/latest/user/configuration.html

sdk/opendp/whitenoise/client/__init__.py
```diff
@@ -45,10 +57,11 @@ def register(self, dataset):
             dataset_type=dataset['dataset_type'], \
             budget=dataset['budget'], \
             csv_details=dataset['csv_details'], \
-            dataverse_details=dataset['dataverse_details'])
+            dataverse_details=dataset['dataverse_details'], \
+            released=False)
         return client.datasetregister(register_request)
 
-    def read(self, dataset_name, budget):
+    def read(self, dataset_name, budget=None):
```
eedeleon: I would prefer a read_release than have budget be optional. 

These are very distict user actions so we should keep them separate. One is used by the analyst the other  is used by the module. Over time the credentials needed for both actions will also be different

sdk/opendp/whitenoise/client/restclient/code-model-v1
```diff
@@ -0,0 +1,1735 @@
+{
```
eedeleon: Seems like a new autorest file. I do not think we need it. We should figure out why it exists though. I saw it last time too

sdk/opendp/whitenoise/client/restclient/models/dataset_document.py
```diff
@@ -27,14 +29,16 @@ class DatasetDocument(Model):
         'dataset_name': {'key': 'dataset_name', 'type': 'str'},
         'dataset_type': {'key': 'dataset_type', 'type': 'str'},
         'budget': {'key': 'budget', 'type': 'float'},
+        'released': {'key': 'released', 'type': 'bool'},
```
eedeleon: I get that they are effectively the same. But it may be worth introducing a ReleaseDatasetDocument instead of having this flag. Overloading budget to budget used vs max budget seems problematic longer term. Especially given the very distinct semantics around both.

What do you think?

service/dataset.py
```diff
@@ -23,17 +23,11 @@
                         },
                         "budget":3.0}}
 
+RELEASED_DATASETS = []
```
eedeleon: We should also add one to start as an example

service/dataset.py
```diff
@@ -23,17 +23,11 @@
...
 
-def read(dataset_request):
-    """Get information needed to load the dataset
-
-    :param info: The dataset to read and budget to use.
-    :type info: dict {"dataset_name": str, "budget":int}
-    :return: A dataset document that contains the type and info of the dataset
-    :rtype: dict{"dataset_type": str, dataset_key: dict}
-    """
+def _read_helper(dataset_request):
```
eedeleon: This should take the dict to read from as input and be wrapped by _read_released, _read.

Pretty sure we are always reading from the private datasets now

service/dataset.py
```diff
@@ -48,23 +42,65 @@ def read(dataset_request):
     except:
         # TODO: Temp fix for testing - Do better cleanup if secret missing
         # dataset["dataset_type"]["token"] = {'name':dataset_name,'value':42}
-        pass 
-    
+        pass
+
+    return dataset
+
+def readreleased(dataset_request):
```
eedeleon: We should add a github issue, or preliminary support her, for clients to create a fake user id guid. This will allow us to better test read released for users before we actually have the authorization code path working. You can pass the guid in as a header for now.

https://stackoverflow.com/questions/29386995/how-to-get-http-headers-in-flask can be used to check for the header. 

This should help us get a jumpstart on how to handle an actual Authorization mechanism.

For example:
generate ids 0-9 for each of the separate clients.
register dataset takes an optional list of valid ids 5-9
read requests fail for 1-4
read released costs budget the first time but not subsequent calls for 0-9

service/dataset.py
```diff
@@ -48,23 +42,65 @@ def read(dataset_request):
...
+
+def release(release_request):
+    dataset_name = release_request["dataset_name"]
+
+    if dataset_name not in DATASETS:
+        abort(401, "Dataset id {} doesn't exist.".format(dataset_name))
+
+    # Check secret here, to make sure this comes from service
+
+    dataset = DATASETS[dataset_name]
```
eedeleon: I like how sleek this looks but I think release will be more complicated.

release is effectively a new dataset.register of a released dataset. The name should be unique among released datasets(like datasets) and the new dataset will probably be created within a module so renaming/exposing a previously private dataset as a released dataset won't work.

Client:
- submits module

Module:
 - reads private data in
 - creates something from it that is differentially private
 - registers a new Dataset, ReleasedDataset, that points to this new type of dataset that costs budget to read but is not private.

This means Modules will need to create a local csv(for testing) or a local csv they upload as a dataset file to dataverse and then register that dataset as a released dataset. There may be other sources we support later but those are the two valid ones now. 

No need to work on uploading a dataverse dataset for this PR, adding a local_csv should suffice

sdk/opendp/whitenoise/client/restclient/models/dataset_document.py
lurosenb: I agree - can totally see DatasetDocument becoming bloated in the longtermwithout separating the two

openapi/swagger.yml
```diff
@@ -50,6 +50,44 @@ paths:
...
+         200:
+           description: Successfully return the dataset information
+           schema:
+            $ref: '#/definitions/ReleaseDatasetDocument'
+  /release:
+    post:
+      operationId: dataset.release
+      tags:
+        - datasetRelease
+      summary: Release a registered dataset
```
eedeleon: Release a differentially private dataset

openapi/swagger.yml
```diff
@@ -50,6 +50,44 @@ paths:
...
+    post:
+      operationId: dataset.release
+      tags:
+        - datasetRelease
+      summary: Release a registered dataset 
+      description: Return the details of the released dataset
+      parameters:
+        - name: release_request
+          in: body
+          description: Request to release a specified dataset
```
eedeleon: Request to make a dataset release

sdk/opendp/whitenoise/client/__init__.py
```diff
@@ -34,24 +52,71 @@ def submit(self, params, uri):
...
-    def register(self, dataset):
+    """
+    A client for registering, reading and releasing differentially private
+    datasets using the opendp-whitenoise service
+    """
+    def __init__(self):
+        # Tag requests with this custom header for now
+        self._guid = _guid_header()
+
+    def readreleased(self, dataset_name, budget):
```
eedeleon: read_released. I haven't figured out keeping python convention for connexion functions, we can maintain it for our client.py functions

sdk/opendp/whitenoise/client/__init__.py
```diff
@@ -34,24 +52,71 @@ def submit(self, params, uri):
...
-    def read(self, dataset_name, budget):
+    def register(self, dataset):
+        """
+        Generates a DatasetDocument and sends it to the service.
+        Requests the registration of this private DatasetDocument
+        Tags the request with Client guid.
+        """
+        client = _get_client()
+        register_request = self._register_release_request_helper(dataset)
+        return client.datasetregister(register_request, custom_headers={'client_guid': self._guid})
```
eedeleon: custom_headers=self._custom_headers might be worth it. And you can add that in the constructor.

I did not expect you to add the guid already :) pleasant surprise

sdk/opendp/whitenoise/client/__init__.py
eedeleon: this one does not need budget(the cost of the read is fixed /defined by the released dataset document.

sdk/opendp/whitenoise/client/__init__.py
```diff
@@ -34,24 +52,71 @@ def submit(self, params, uri):
...
+
+    def read(self, dataset_name):
```
eedeleon: We do need budget here for the read request on private data. This will be called from modules

sdk/opendp/whitenoise/client/__init__.py
```diff
@@ -34,24 +52,71 @@ def submit(self, params, uri):
...
             budget=dataset['budget'], \
             csv_details=dataset['csv_details'], \
-            dataverse_details=dataset['dataverse_details'])
-        return client.datasetregister(register_request)
+            dataverse_details=dataset['dataverse_details'], \
+            authorized_users=dataset['authorized_users'])
+        
+        return request
+
+    def release(self, dataset):
```
eedeleon: the dataset document should include a budget param that is the cost of creating that release

service/modules/dataset-module/MLproject
```diff
@@ -0,0 +1,11 @@
+name: diffpriv_sql
+
+conda_env: conda.yaml
+
+entry_points:
+  main:
+    parameters:
+      dataset_name: str
+      budget: double
+      query: str
```
eedeleon: query -> released dataset name

service/modules/dataset-module/register_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+    with mlflow.start_run():
+        new_dataset = {
+            "dataset_name": dataset_name,
+            "dataset_type": dataset_type,
+            dataset_type: details,
+            "budget": budget,
+            "authorized_users": []
+        }
+
+        dataset_document = get_dataset_client().register(new_dataset)
```
eedeleon: release_dataset(new_dataset)

service/modules/dataset-module/register_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+    # Else, load each specified detail
+    if json_flag == "True":
+        with open(sys.argv[5], "r") as f:
+            details = json.load(f)
+    else:
+        # Args split by spaces, then key=value
+        for arg in sys.argv[5:]:
+            details[arg.split("=")[0]] = arg.split("=")[1]
+    
+    with mlflow.start_run():
```
eedeleon: We need to create a result.json with the correct released_dataset name

service/modules/dataset-module/register_dataset.py
```diff
@@ -0,0 +1,48 @@
...
+
+"""
+Sample Command line:
+
+python register_dataset.py private_csv csv_details 10.0 False local_path=serverside/path/to/example.csv
+"""
+
+if __name__ == "__main__":
+    dataset_name = sys.argv[1]
+    dataset_type = sys.argv[2]
```
eedeleon: The module:

dataset = load_dataset(get_dataset(dataset_name))

do stuff

For csv details:  Write your new dataset to os.path.join(os.path.dirname(csv_details path) ,released_dataset_name.csv)
Copy the previous schema to the os.path.join(os.path.dirname(csv_details path) ,released_dataset_name.yaml)

Create the new details (no passing it in the param)
Set the budget on the new details
create a result.json("released_dataset_name":"name")
register_released(new_details)

service/dataset.py
```diff
@@ -21,25 +26,46 @@
...
+KNOWN_USERS = {'ab30c2f7-d97b-42b5-9e74-f9ed05b43839': defaultdict(int), 'mock_creds': defaultdict(int, {'example_released_csv': 1})}
 
-    :param info: The dataset to read and budget to use.
-    :type info: dict {"dataset_name": str, "budget":int}
-    :return: A dataset document that contains the type and info of the dataset
-    :rtype: dict{"dataset_type": str, dataset_key: dict}
-    """
+# NOTE: This could be tracked inside datasets as well, although
+# potentially safer to only track owners service side
+OWNERS = {'ab30c2f7-d97b-42b5-9e74-f9ed05b43839': ["example_csv"]}
```
eedeleon: We may need owners longer term but for now an owner can be marked as having read the dataset, and therefore not incur future costs

service/dataset.py
```diff
@@ -48,60 +74,165 @@ def read(dataset_request):
...
+        abort(404, "User not authorized to access this dataset.")
+
     # Check/Decrement the budget before returning dataset
-    adjusted_budget = dataset["budget"] - dataset_request["budget"]
-    if adjusted_budget >= 0.0:
-        dataset["budget"] = adjusted_budget
-    else:
-        abort(412, "Not enough budget for read. Remaining budget: {}".format(dataset_name))
+    # Note: We are guaranteed that client_guid in KNOWN_USERS
+    if client_guid not in KNOWN_USERS:
```
eedeleon: From our chat I thought we landed on a simpler approach. No authorized users yet(that was a possible future thing)

For now we have USERS = {}
release_dataset(origi_dataset, budget, release_dataset_name) -> adds the owner to USERS_RELEASED[dataset_name][owner_id] += 1)
read_released_dataset(released_dataset_name) USERS_RELEASED[dataset_name][user_id] +=1 if USERS_RELEASED[dataset_name][user_id] > 1 charge budget

We can do a similar thing for authorized users if you want to add it where first we verify user_id in AUTHORIZED_USERS[released_dataset_name]  if not fail

eedeleon: Budget is global so there is no need to block for new users since we are not separating budget between authorized users and the global pool.

sdk/opendp/whitenoise/client/__init__.py
lurosenb: I added a new field called "release_cost," although should there be bothe a general "budget" field and a "release_cost" field? (both are currently in the spec)

eedeleon: We can leave them in one doc for now.

service/dataset.py
```diff
@@ -21,25 +26,48 @@
...
+                            "schema": '{"fake":"schema"}'
+                        },
+                        "budget":10.0,
+                        "authorized_users":['mock_creds']}}
 
 KNOWN_DATASET_TYPE_KEYS = ["csv_details", "dataverse_details"]
 
-def read(dataset_request):
-    """Get information needed to load the dataset
+KNOWN_USERS = {'ab30c2f7-d97b-42b5-9e74-f9ed05b43839': defaultdict(int), 'mock_creds': defaultdict(int, {'example_released_csv': 1})}
```
eedeleon: >>> int_default_dict = lambda : defaultdict(int)
>>> my_dict = defaultdict(int_default_dict)
>>> my_dict["hi"]["hi"] += 1
>>> my_dict
defaultdict(<function <lambda> at 0x7fab9e16e840>, {'hi': defaultdict(<class 'int'>, {'hi': 1})})

service/dataset.py
```diff
@@ -48,60 +76,151 @@ def read(dataset_request):
...
+    # Checks to see if authorized users specified
+    if 'authorized_users' in dataset and len(dataset['authorized_users']) > 0:
+        if client_guid not in dataset['authorized_users']:
+            abort(404, "User not authorized to access this dataset.")
+
+    # Track readrelease
+    USERS_RELEASED[dataset_name][client_guid] += 1
+
+    # Decrement budget if this is a users first read
+    if USERS_RELEASED[dataset_name][client_guid] > 1:
```
eedeleon: I don't think we need this section until we have per user budget. With global budget the cost of the dataset was used. We can keep user incrementing since it will be important later

eedeleon: Also this counts as state so we would like to avoid it, we will probably have to do an accounting document for user reads of released datasets longer term. This is good for our reference implementation

service/dataset.py
```diff
@@ -48,60 +76,151 @@ def read(dataset_request):
...
-    # TODO: Add support for other types of datasets
+        if not dataset_request["dataverse_details"]["host"]:
+            abort(408, "Must specify host, {} is malformed.".format(str(dataset_request["dataverse_details"]["host"])))
 
+    # Track owner
+    client_guid = request.headers.get('client_guid')
+    if client_guid in OWNERS:
+        OWNERS[client_guid].append(dataset_name)
+    else:
+        OWNERS[client_guid] = [dataset_name]
```
eedeleon: default dict list?

service/dataset.py
```diff
@@ -48,60 +76,151 @@ def read(dataset_request):
...
+        abort(401, "Dataset id {} already exists. Identifiers must be unique".format(dataset_name))
+
+    # Check key
+    if dataset_request["dataset_type"] not in KNOWN_DATASET_TYPE_KEYS:
+        abort(402, "Given type was {}, must be either csv_details or dataverse_details.".format(str(dataset_request["dataset_type"])))
     
-    # Add budget if possible 
-    if dataset["budget"]:
-        if dataset["budget"] <= 0.0: abort(403, "Budget must be greater than 0.")
+    # Check budget
```
eedeleon: You mentioned release_budget? Where is that set?

openapi/swagger.yml
lurosenb: resolved

sdk/opendp/whitenoise/client/__init__.py
lurosenb: will add flake in future PR

sdk/opendp/whitenoise/client/restclient/code-model-v1
lurosenb: removed, lol

service/dataset.py
lurosenb: example added everywhere

service/dataset.py
lurosenb: nice system test added with this scenario in tests/dataset/test_smoke.py

service/dataset.py
lurosenb: Module has been updated according to this spec

service/modules/dataset-module/register_dataset.py
lurosenb: module has been updated according to this spec

service/dataset.py
lurosenb: dope, and added

service/dataset.py
lurosenb: removed for now - will have to add user accounting in future PR

service/dataset.py
lurosenb: only set by document owner, handled in release

service/datasets/iris.csv
```diff
@@ -0,0 +1,151 @@
+,sepal length (cm),sepal width (cm),petal length (cm),petal width (cm)
```
eedeleon: Add this to the git ignore and remove it from the PR please.

service/modules/dataset-module/conda.yaml
```diff
@@ -0,0 +1,9 @@
+name: tutorial
```
eedeleon: We should fix the import. You can use opendp-whitenoise now that we are on pypi

service/modules/dataset-module/README.md
```diff
@@ -0,0 +1 @@
+# Partial Differentiall Private SQL module Mlflow project
```
eedeleon: This needs to be updated for dataset

service/modules/dataset-module/release_dataset.py
```diff
@@ -0,0 +1,56 @@
...
+            "local_csv": {},
+            "budget": budget,
+            "authorized_users": []
+        }
+
+        with open("result.json", "w") as stream:
+            json.dump({"released_dataset_name": release_dataset_name}, stream)
+
+        # TODO: Perform basic dataset_document validation
+        service_client.release(new_dataset)
```
eedeleon: The mlflow artifact needs to be created for result.json

PR #171, opened 2020-03-30
DPSU Module

anujdesai510:


tests/service/execute/test_dpsu.py
```diff
@@ -0,0 +1,8 @@
+import pytest
+
+@pytest.mark.parametrize("project", [{"params": {"dataset_name": "PUMS", "budget": .2,
+                                      "query": "SELECT married, educ FROM PUMS.PUMS GROUP BY married, educ"},
```
eedeleon: A local test, not in execution with private reader should run, and make sure that it is using dpsu if the right params are set

eedeleon: Maybe adding an assert at the end would be enough to verify it.

assert PrivateReader()._get_reader is not PrivateReader.reader

And maybe a negative test if the reader is not a PandasReader

sdk/opendp/whitenoise/sql/dpsu.py
```diff
@@ -5,19 +5,22 @@
...
     """
-    queries = QueryParser(schema).queries(query)
+    qp = QueryParser(schema)
+    q = qp.query(query_string)
+    queries = qp.queries(query_string)
     query_ast = queries[0]
 
     group_cols = [ge.expression.name for ge in query_ast.agg.groupingExpressions]
-    table_name = query_ast.m_sym_dict[group_cols[0]].tablename
+    table_name = q.source.find_node(Table).name
```
eedeleon: That is so clean

PR #172, opened 2020-04-02
Feat synthetic data

lurosenb:
Synthetic data module
- Added new synthetic module inside service
- Dependency on sdgym PrivBayes (only currently supported synthetic data mechanism)
   NOTE: sdgym PyPi does not ship with PrivBayes, as it has a dependency on a c binary
   (So needed to copy over those files)


service/modules/synthetic-data-module/MLproject
```diff
@@ -0,0 +1,11 @@
...
+
+conda_env: conda.yaml
+
+entry_points:
+  main:
+    parameters:
+      dataset_name: str
+      budget: double
+      query: str released dataset name
+    command: "python run_query.py {dataset_name} {budget} {query}"
```
eedeleon, line 11: query -> release_dataset_name

service/modules/synthetic-data-module/README.md
```diff
@@ -0,0 +1,22 @@
+# Produce synthetic data from a document, evaluate with sdgym
+
+## Installing sdgym
+
+### Pip install
+
+Note: Make sure you have Cython installed, as well as scikit learn.
```
eedeleon, line 7: We should have a Dockerfile for this soon. I may ping you on what would be needed for it

service/modules/synthetic-data-module/privbayes/code/printer.cpp
```diff
@@ -0,0 +1,60 @@
+#include "printer.h"
+
```
eedeleon, line 2: Not sure we can check this in since it is code from another OSS library. Maybe we can point to the git url for the raw files and download them in the module?

service/modules/synthetic-data-module/synthetic_data.py
```diff
@@ -0,0 +1,87 @@
...
+        
+        # TODO: Add check to validate dataset.to_numpy
+        synthesizer.fit(dataset.to_numpy(), synth_schema['categorical_columns'], synth_schema['categorical_columns'])
+        synthetic_data = synthesizer.sample(int(sample_size))
+
+        # NOTE: Two options here.
+        # 1. CURRENTLY IMPLEMENTED: Can default return the synthetic data in the payload.
+        # 2. Can default release a new dataset, and return name.
+        
+        df = pd.DataFrame(synthetic_data,
```
eedeleon, line 82: For now we can only support local files + local service. 

Also we should release the synthetic data df, as a new local_csv dataset, not return it in the response

eedeleon: os.dirname(dataset_document.local_csv_path)/released_dataset_name.csv
cp dataset_document.local_csv_schema   -> os.dirname(dataset_document.local_csv_path)/released_dataset_name.yaml

PR #173, opened 2020-04-05
validate dpsu with reddit data

anujdesai510:


tests/conftest.py
```diff
@@ -45,6 +45,53 @@
     schema = CollectionMetadata([iris], "csv")
     schema.to_file(iris_schema_path, "iris")
 
+def find_ngrams(input_list, n):
+    if n == 1:
```
eedeleon, line 49: return input_list if n ==1 else list(zip(*[input_list[i:] for i in range(n)]))

PR #176, opened 2020-04-08
Synthetic Data Module + MWEM

lurosenb:
- Added MWEM synthesizer implementation
  - Python implementation based on numpy ndarray histograms of datasets
  - Potential improvements include improving performance on higher dimension data
- Added synthetic data module (uses sdgym, a new dependency)
  - Allows for the creation of DP synthetic data through the service
  - Only supports MWEM as of now, TODOs: include adding support for other synthesizers


eedeleon:
Please squash merge the changes

PR #192, opened 2020-04-27
Adding pkgutil to share the `opendp.whitenoise` namespace

raprasad:
e.g. Namespace to be shared with packages generated from https://github.com/opendifferentialprivacy/whitenoise-core-python


raprasad:
resolved

PR #193, opened 2020-05-04
Feature: Update MWEM

lurosenb:
Functionality:
Added ability to factor histograms, so that MWEM scales with any size dataset (no longer limited by dimensionality)

Architecture:
Moved MWEM into SDK, where it seems to fit more naturally. Added unit tests.


service/modules/synthetic-data-module/synthetic_data.py
```diff
@@ -16,7 +16,7 @@
 from sdgym.constants import CONTINUOUS
 from sdgym.synthesizers.utils import Transformer
 
-from mwem.mwem import MWEMSynthesizer
+from opendp.whitenoise.synthesizers.mwem.mwem import MWEMSynthesizer
```
eedeleon, line 19: Can this be from opendp.whitenoise.synthesizers.mwem import MWEMSynthesizer.

Also, in 
synthesizers/__init__.py, you can add a shorter import like we do in sql/__init__.py

sdk/opendp/whitenoise/synthesizers/mwem/mwem.py
```diff
@@ -6,9 +7,7 @@
 import numpy as np
 import pandas as pd
 
-from sdgym.synthesizers.base import BaseSynthesizer
-
-class MWEMSynthesizer(BaseSynthesizer):
```
eedeleon, line 11: Can we have an SDGYMBaseSynthesizer class that defines the minimal spec in synthesizers/base.py. 

It can be an abstract class to more clearly guide developers on what functions are needed for synthetic data generation.

If SDGYM were 1.0 we would depend on it directly, for now we can start off work like this. And then reach out to see if the collaboration can be more close.

sdk/opendp/whitenoise/synthesizers/mwem/mwem.py
```diff
@@ -23,15 +22,17 @@ class MWEMSynthesizer(BaseSynthesizer):
...
         self.iterations = iterations
         self.mult_weights_iterations = mult_weights_iterations
         self.synthetic_data = None
         self.data_bins = None
         self.real_data = None
-
+        self.splits = splits
+        self.split_factor = split_factor
+        
     def fit(self, data, categorical_columns=tuple(), ordinal_columns=tuple()):
```
eedeleon, line 36: Should the two optional params  default to None or tuple?

sdk/opendp/whitenoise/synthesizers/mwem/mwem.py
```diff
@@ -50,6 +51,9 @@ def fit(self, data, categorical_columns=tuple(), ordinal_columns=tuple()):
             self.data = data.copy()
         else:
             raise ValueError("Data must be a numpy array.")
+
+        if self.split_factor > 0 and self.splits == []:
```
eedeleon, line 55: Do we use split_factor as -1 or mostly check > 0 and avoid it? If so, it would be better to have None to not have unintended bugs

sdk/opendp/whitenoise/synthesizers/mwem/mwem.py
```diff
@@ -150,7 +178,7 @@ def initialize_A(self, histogram, dimensions):
         A += value
         return A
 
-    def histogram_from_data_attributes(self, data):
+    def histogram_from_data_attributes(self, data, splits=[]):
```
eedeleon, line 181: Do we need to expose the non 
.fit
.sample

functions? If not, can we make them _histogram_from_data_attributes(...). That way we can keep the complexity away from the users

sdk/opendp/whitenoise/synthesizers/mwem/mwem.py
lurosenb: Gotcha, added

sdk/opendp/whitenoise/synthesizers/mwem/mwem.py
lurosenb: good catch, meant to remove these

sdk/opendp/whitenoise/synthesizers/mwem/mwem.py
lurosenb: good call

sdk/opendp/whitenoise/synthesizers/mwem/mwem.py
lurosenb: good idea

Issue #209, opened 2020-05-23
SQL queries without a "GROUP BY" clause have a weird behavior


TedTed:
Not critical, just strange: trying to run an SQL query without any GROUP BY returns [only one row](https://twitter.com/TedOnPrivacy/status/1263546652645830659), with very strange results. Not sure what's happening here, but the query should probably fail or return zero rows rather than nonsense.

joshua-oss:
Thanks for flagging this.  We removed a bunch of checks when we decided the analyst would be trusted, but this behavior is definitely wrong.  Adding some checks back in.

joshua-oss:
PR #210 adds check to throw error when analyst runs query with missing aggregation on GROUP BY

joshua-oss:
This is fixed

Issue #201, opened 2020-05-21
The SQL PrivateReader does clamping too early


TedTed:
Currently, the "clamping" only happens at the lowest level: when a table column is called. This means that I can do [attacks like these](https://twitter.com/TedOnPrivacy/status/1263542430089056261) that can trivially leak the data of a user. This is particularly adversarial, but someone could use logic like that without meaning to run an attack (simply because it's convenient) and end up with very different privacy guarantees than what they expect, if any.

Instead, the clamping needs to happen on the values passed to the aggregation. To do that, you need the user to set clamping parameters *in the query*, or you need a way of automatically estimating these bounds (in a DP way), like what is described in [this paper](https://arxiv.org/abs/1909.01917). In the meantime, queries that do anything else than aggregating a know column with metadata should probably be forbidden.

joshua-oss:
Thanks for catching this.  The validator should be prohibiting anything other than simple column select inside aggregations.  We will investigate and fix.

joshua-oss:
PR #210 now prevents analysts from performing expressions inside of aggregations.  Computing sensitivity and stability inside aggregations is not always possible, and something we will delegate to the analysis graph in the future.

joshua-oss:
This has been fixed

PR #210, opened 2020-05-24
Add back some validator checks

joshua-oss:
Adding back some validator checks we removed under assumption analyst is trusted.  There are some obvious cases where the analyst could inadvertently harm privacy, so attempt to enforce:

* Aggregates only allowed over Columns (no expressions)
* GROUP BY only allowed over columns
* Simple arithmetic allowed as post-processing
* No combination of expressions in post-processing, even if it's safe
* No JOINS or Subqueries anywhere

This eliminates a class of queries that are safe, which were previously supported.  I've deleted those previously passing unit tests, and we can change documentation to say data curator should preprocess in a materialized view is data is known-safe.


joshua-oss:
Related #201, #209

sdk/opendp/whitenoise/ast/validate.py
```diff
@@ -53,22 +86,35 @@ def check_groupkey(self):
         if (len(gbk) > 0) and (len(gbk) == len(gc)):
             raise ValueError("GROUP BY must include more than key columns: " + ", ".join([str(g) for g in gc]))
 
+    def check_select_relations(self):
+        rel_nodes = self.query.select.find_nodes(SqlRel)
+        if (len(rel_nodes)) > 0:
+            raise ValueError("We don't support subqueries in the SELECT clause")
+
+    def check_order_relations(self):
```
eedeleon, line 94: I could not find any use of either of these two functions. Are they going to be used in a follow up PR?

sdk/opendp/whitenoise/ast/validate.py
```diff
@@ -37,13 +37,46 @@ def validate_all(self):
...
+                    exp = r
+                elif type(r) is Literal:
+                    exp = l
+            if isinstance(exp, (PowerFunction, RoundFunction)):
+                exp = exp.expression
+            return exp
+
+        select_expressions = [simplify(s[1]) for s in q.m_symbols]
+        is_agg = lambda n: type(n) == AggFunction and n.is_aggregate
+        split = lambda p, i: (list(filter(p, i)), list(filter(lambda v: not p(v), i)))
```
eedeleon, line 59: I think python supports this better with list comprehensions. Although the functional approach is very clean and I was able to follow it easily. filter in python is inferior to list comprehensions since it requires the lambda for not where you can do the below code with list comprehensions and remove the need for a split function.

aggs = [val if is_agg(val) for val in select_expressions]
non_aggs = [val if not is_agg(val) for val in select_expressions]

joshua-oss: Good call.  In functional languages, it's a single-line command that splits an iterable based on a single predicate, and it makes me nervous that python forces me to copy the predicate reference in two places.  But my "fix" is worse than the problem.  I'll change it.

sdk/opendp/whitenoise/ast/validate.py
joshua-oss: Oops, I decided not to validate ORDER BY, since it's done in post-processing and has no impact on rewriter or privacy.  I've removed this check now.

joshua-oss:
> I have some open questions around simplify. Is it never the case that it could omit nodes? for example if the last left and right node are both expressions? IS that ok?

The 'simplify' routine eliminates nodes that don't matter for purposes of validation, but potentially changes the semantics of the query, so we wouldn't want to actually run that query.  For example, something like (((SUM(Sales))) / ((10.0 * 10.0))) could be simplified to SUM(Sales) for purposes of validation, since constant transforms in post-processing have no impact on privacy.  However, if we ran the 'simplified' query, it would return a value that is 100X too large.

The AST used by the validator is built by the validator internally and then thrown away, so we aren't manipulating the caller's query.

Issue #204, opened 2020-05-21
The SQL thresholding math seems off: tau uses a Laplace formula, but the noise is Gaussian


TedTed:
Looking at `private_reader.py`, it seems that you're using [Gaussian](https://github.com/opendifferentialprivacy/whitenoise-system/blob/60743c0a6bd2002296764a5c388e999c2b99a7dd/sdk/opendp/whitenoise/sql/private_reader.py#L156) as a default mechanism, but the threshold `tau` is computed [based on a Laplace formula](https://github.com/opendifferentialprivacy/whitenoise-system/blob/60743c0a6bd2002296764a5c388e999c2b99a7dd/sdk/opendp/whitenoise/sql/private_reader.py#L123), so it's probably wrong.

This file also imports Laplace but doesn't use it. This should probably be caught by a linter.

joshua-oss:
This is fixed now in #211, merged to master

joshua-oss:
This is fixed now

PR #211, opened 2020-05-25
Use Gaussian thresholds for counts

joshua-oss:
Code for threshold calculation got out of sync with mechanism being used.  This will not be a problem when we add stability histogram API in whitenoise-core mechanisms.  Updated manually now and added unit tests that will catch if this happens again.

### DIFF
--- a/tests/sdk/query/test_query.py
+++ b/tests/sdk/query/test_query.py
@@ -3,6 +3,7 @@ import subprocess
 
 import pandas as pd
 from pandasql import sqldf
+import math
 
 from opendp.whitenoise.metadata import CollectionMetadata
 from opendp.whitenoise.sql import PrivateReader, PandasReader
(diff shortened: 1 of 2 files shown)


joshua-oss:
Fix for #204, pending stability histogram API surface on core

tests/sdk/query/test_query.py
```diff
@@ -131,3 +132,23 @@ def test_execute_without_dpsu(self):
         query = QueryParser(schema).queries("SELECT COUNT(*) AS c FROM PUMS.PUMS GROUP BY married")[0]
         private_reader.options.use_dpsu = False
         assert(private_reader._get_reader(query) is private_reader.reader)
+    def test_check_thresholds_gauss(self):
```
eedeleon, line 135: did this test fail before the proposed changes?

PR #220, opened 2020-05-29
Cleaning comments in Evaluation

ankit-oss:
Get comments consistent below methods and classes for clarity
Remove a few unnecessary print statements


sdk/opendp/whitenoise/evaluation/_exploration.py
```diff
@@ -77,16 +95,12 @@ def generate_powerset(self, d1):
...
             self.generate_powerset(df)
-            print(self.visited)
-
-    def main(self):
-        self.test_exploration()
-
-if __name__ == "__main__":
-    ex = Exploration()
-    ex.main()
+            print(self.visited)
```
eedeleon, line 106: Should this print be here?

ankit-oss: Good catch Eddie! Removed it.

PR #222, opened 2020-05-29
dpgan synthesizer

anujdesai510:
API cleanup:
- Preprocessor class to convert from pandas df to numpy array that can be fed as input into GAN
- GAN class to train with torchdp and generate samples
- PytorchDPSynthesizer wrapper class that implements SDGYMBaseSynthesizer spec that constructs a synthesizer by specifying a Preprocessor and GAN


sdk/opendp/whitenoise/synthesizers/dpgan/dpgan.py
```diff
@@ -0,0 +1,184 @@
...
+        self.batch_size = batch_size
+        self.epochs = epochs
+        self.n_critics = n_critics
+        self.delta = delta
+        self.budget = budget
+        
+        self.device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+    
+    def fit(self, data, categorical_columns=tuple(), ordinal_columns=tuple()):
+        self.transformer = GeneralTransformer()
```
anujdesai510, line 92: @eedeleon this line here is what I'm referring to. Here we instantiate the object that creates a vector embedding from the original pandas dataframe which then gets passed to the discriminator network. The vector embedding might change though if someone uses an autoencoder or some other preprocessing step for example

sdk/opendp/whitenoise/synthesizers/dpgan/dpgan.py
```diff
@@ -0,0 +1,184 @@
...
+            for i, data in enumerate(dataloader):
+                discriminator.zero_grad()
+                
+                real_data = data[0].to(self.device)
+                
+                # train with fake data
+                noise = torch.randn(self.batch_size, self.latent_dim, 1, 1)
+                noise = noise.view(-1, self.latent_dim)
+                fake_data = self.generator(noise)
+                fake_data = apply_activate(fake_data, self.transformer.output_info)
```
anujdesai510, line 133: The issue becomes here, where we want to attach a different activation based on the datatype: softmax for categorical, tanh for continuous

sdk/opendp/whitenoise/synthesizers/pytorch/nn.py
```diff
@@ -0,0 +1,155 @@
...
+
+                if self.delta is None:
+                    self.delta = 1 / data.shape[0]
+                
+                epsilon, best_alpha = optimizer_d.privacy_engine.get_privacy_spent(self.delta)
+                
+            if self.budget < epsilon:
+                break
+
+    def generate(self, n):
```
eedeleon, line 142: Is this the typical GAN API?

anujdesai510: Some people optionally include a generate method but in this case in order for sample() to be agnostic to PyTorch we need one. Besides that, the train() is the same but with torchdp. We could also make that optional and only create a privacy engine if privacy=True

sdk/opendp/whitenoise/synthesizers/pytorch/nn.py
```diff
@@ -0,0 +1,160 @@
...
+    import torch
+    import torch.nn as nn
+    import torch.nn.functional as F
+    import torch.optim as optim
+    from torch.utils.data import DataLoader, TensorDataset
+
+    from torchdp import PrivacyEngine, utils, autograd_grad_sample
+except ImportError:
+    import logging
+    test_logger = logging.getLogger(__name__)
```
eedeleon, line 13: These try catches should be in the tests, not the library

tests/sdk/synthesizers/test_pytorch_synthesizer.py
```diff
@@ -0,0 +1,34 @@
+import subprocess
+import os
+
+import pytest
+import string
+import pandas as pd
+
+from opendp.whitenoise.metadata import CollectionMetadata
+from opendp.whitenoise.synthesizers.preprocessors.preprocessing import GeneralTransformer
+from opendp.whitenoise.synthesizers.pytorch.pytorch_synthesizer import PytorchDPSynthesizer
```
eedeleon, line 10: We should try catch this line for the builds

tests/sdk/synthesizers/test_pytorch_synthesizer.py
```diff
@@ -0,0 +1,50 @@
...
+            self.dpgan = PytorchDPSynthesizer(GeneralTransformer(), DPGAN())
+        except:
+            raise Exception()
+
+    def test_fit(self):
+        try:
+            self.dpgan.fit(df)
+            assert self.dpgan.gan.generator
+        except:
+            raise Exception()
```
eedeleon, line 41: remove the try excepts from this file

Issue #216, opened 2020-05-28
Create Privacy Algorithm interface


ankit-oss:
A lot of the heavy lifting done by aggregation.py in evaluator needs to be taken up by privacy algorithms themselves for the evaluator SDK to become generic. So, the evaluator SDK shall provide a set of interfaces that the privacy algorithms can implement and run the stochastic tests on the data they like and the queries that they want to test.

PR #229, opened 2020-06-03
Evaluator Base Interface

ankit-oss:
Defined base class interfaces for building evaluator as a generic implementation for black box DP implementations to leverage

Created a sample implementation of BlackBoxPrivacy Interface and tested it

Sample implementation of a DP library that releases statistics by implementing BlackBoxPrivacyInterface. It takes in the privacy and evaluation params for its analysis and releases noisy responses for evaluator to act on.

**Test Result**
(oss_dp) ankit@ANKIT-DELL:~/privacytoolsproject/whitenoise-system/tests$ pytest -k "test_dp_blackbox"
================================ test session starts ===========================
platform linux -- Python 3.8.0, pytest-5.3.1, py-1.8.0, pluggy-0.13.1
rootdir: /mnt/c/Users/ankitsri/source/code/privacytoolsproject/whitenoise-system/tests, inifile: pytest.ini
collected 116 items / 115 deselected / 1 selected                                                                                         

sdk/evaluation/test_interface.py .                                                                                                  [100%]

=========================== 1 passed, 115 deselected in 3.41s ========================= 

Related #216

### DIFF
--- a/sdk/opendp/whitenoise/evaluation/blackbox/_base.py
+++ b/sdk/opendp/whitenoise/evaluation/blackbox/_base.py
@@ -15,6 +15,7 @@ class BlackBoxPrivacyInterface:
 		it can be a graph, some sort of script written in any language, or a SQL
 		query. privacy_params are a shared format that is consumed by the evaluator
 		"""
+		pass
 	
 	def release(self, dataset : object, actual = False) -> Report:
 		"""
@@ -25,3 +26,4 @@ class BlackBoxPrivacyInterface:
 		
 		Returns reports as a in-memory Python tabular vector of arrays.
 		"""
+		pass
--- a/sdk/opendp/whitenoise/evaluation/blackbox/_base.py
+++ b/sdk/opendp/whitenoise/evaluation/blackbox/_base.py
@@ -1,13 +1,15 @@
 from opendp.whitenoise.evaluation.params._privacy_params import PrivacyParams
 from opendp.whitenoise.evaluation.params._eval_params import EvaluatorParams
 from opendp.whitenoise.evaluation.report._report import Report
+from abc import ABC, abstractmethod
 
-class BlackBoxPrivacyInterface:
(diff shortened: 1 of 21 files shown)


sdk/opendp/whitenoise/evaluation/blackbox/_base.py
```diff
@@ -0,0 +1,32 @@
+from opendp.whitenoise.evaluation.params._privacy_params import PrivacyParams
+from opendp.whitenoise.evaluation.params._eval_params import EvaluatorParams
+from opendp.whitenoise.evaluation.report._report import Report
+from abc import ABC, abstractmethod
+
+class BlackBoxPrivacyInterface(ABC):
```
eedeleon, line 6: why is this internal?

eedeleon: nvm, all of it is so we can test it

ankit-oss: Ya, I really liked the internal convention which allows me to check-in and then we can make it external when I implement the interfaces and add unit tests that pass. So, for skeleton, keeping it internal for now.

sdk/opendp/whitenoise/evaluation/report/_report.py
```diff
@@ -0,0 +1,33 @@
+import pandas as pd
+import numpy as np
+
+class Report:
```
eedeleon, line 4: How does this differ from the core library report?

ankit-oss: That's a good question. This one is agnostic to any specific DP implementation. All it needs is to pass the result of a PrivateQuery or DP function as a dataframe that we can then use for doing our evaluator checks. So quite lightweight. It allows for multiple dimensions (i.e. group by fields) and numerical columns within output to be tested in one-go.

ankit-oss:
Thanks Eddie!

PR #234, opened 2020-06-05
Powerset test switch to use DP group by test

ankit-oss:
Powerset was so far using the dp_query_test that only works for singleton queries. This PR changes it to use the GROUP BY test which can work on noisy responses that have more than 1 row i.e. have dimension keys along with numerical responses.

Please note GROUP BY test is returning no rows when we have just 3 records like in the case of a halton sample. This is happening even after censor_dims is set to False. So, we'll have to find a better approach to run powerset test on GROUP BY queries that run on 3 record datasets.


sdk/opendp/whitenoise/evaluation/_aggregation.py
```diff
@@ -102,17 +103,20 @@ def run_agg_query(self, df, metadata, query, confidence, get_exact=True):
...
         # Getting exact result
         reader = PandasReader(metadata, df)
         exact = reader.execute_typed(query).rows()[1:]
         exact_res = []
         for row in exact:
             exact_res.append(row)
 
-        private_reader = PrivateReader(metadata, reader, self.epsilon)
+        prOptions = PrivateReaderOptions(censor_dims = False)
+        private_reader = PrivateReader(metadata, reader, self.epsilon, options=prOptions)
```
eedeleon, line 119: naming convention should be snake case pr_options

eedeleon:
Deprecated code path, the PR is not longer needed in this form

PR #248, opened 2020-07-16
add dpctgan in sdk 

AprilXiaoyanLiu:
- convert nn.py to a module 
- copy nn.py and paste to __init__.py in nn 
- add dpctgan.py under nn folder
- dpctgan leverages its own transformer (though very similar to transformer implemented in preprocessor.py), so edit code in pytorch_synthesizer to account for the situation when preprocessor is None and situation that doesn't need extra inverse_transformer step.  Please let me know which is preferred. I can also make updates to separate preprossor and inserse_transformer step like what dpgan implemented. the reason I didn't separate the preprocessor step is because during training step, it needs call self.transformer.output_dim and other attribute of transformer. 
- since dpctgan transform data in training step, pass categorical_column and ordinal_columns as argument when calling self.gan.train() 
------------------------------
UPDATED IN 8/4/2020

- remove Monkeypatch of dpctgan, replace it with Discriminator class
- update pytorchsnthesizer to handle preprocessor as None scenario 
- update pytorchsynthesizer to handle generated output as array or dataframe 

@eedeleon  @anujdesai510 @lurosenb


sdk/opendp/whitenoise/synthesizers/pytorch/nn/dpctgan.py
```diff
@@ -0,0 +1,331 @@
...
+            seq += [Linear(dim, item), LeakyReLU(0.2), Dropout(0.5)]
+            dim = item
+
+        seq += [Linear(dim, 1)]
+        if loss == 'cross_entropy':
+            print ("this should happen")
+            seq += [Sigmoid()]
+        self.seq = Sequential(*seq)
+        
+Discriminator.__init__ = __init__
```
eedeleon, line 45: Why do we do this?

sdk/opendp/whitenoise/synthesizers/pytorch/nn/dpctgan.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+
+
+
+
+
+class DPCTGAN(SDGYMBaseSynthesizer, CTGANSynthesizer):
```
eedeleon, line 53: We should not need this class anymore. Or at least, it should not have the Synthesizers. The GAN should be constructed from here and passed to PytorchDPSynthesizer

sdk/opendp/whitenoise/synthesizers/pytorch/nn/dpctgan.py
AprilXiaoyanLiu: because we need add Sigmoid() layer for cross entropy loss,  the default loss function that CTGAN used is wasserstein loss

AprilXiaoyanLiu: > Why do we do this?
because we need add Sigmoid() layer for cross entropy loss, the default loss function that CTGAN used is wasserstein loss

sdk/opendp/whitenoise/synthesizers/pytorch/nn/dpctgan.py
AprilXiaoyanLiu: oh, right, SDGYMBaseSynthesizer isn't needed here.  I think CTGANSynthesizer is still needed, because DPCTGAN is extending the CTGANSynthesizer import from ctgan.

sdk/opendp/whitenoise/synthesizers/pytorch/nn/__init__.py
```diff
@@ -67,7 +74,7 @@ def __init__(self,
         self.pd_cols = None
         self.pd_index = None
 
-    def train(self, data):
+    def train(self, data, categorical_columns=tuple(), ordinal_columns=tuple()):
```
anujdesai510, line 77: We shouldn't need categorical_columns and ordinal_columns as arguments here. They're only used to indicate how preprocessing gets handled but the GAN classes should simply take a matrix that has already been transformed as input to train()

sdk/opendp/whitenoise/synthesizers/pytorch/nn/dpctgan.py
anujdesai510: is the discriminator from CTGAN different than the one from nn/__init__.py? We should be able to use Generator and Discriminator across GAN architectures and Discriminator contains an arg "wasserstein" that indicates whether to add a Sigmoid layer when false

sdk/opendp/whitenoise/synthesizers/pytorch/nn/dpctgan.py
```diff
@@ -0,0 +1,331 @@
...
+        self.verbose=verbose
+        self.loss=loss
+
+
+    def train(self, train_data, categorical_columns=tuple(), ordinal_columns=tuple()):
+
+        
+        self.transformer = DataTransformer()
+        self.transformer.fit(train_data, categorical_columns)
+        train_data = self.transformer.transform(train_data)
```
anujdesai510, line 107: You should be able to remove this and call PytorchDPSynthesizer(DataTransformer(), DPCTGAN()), where the output of DataTransformer is the numpy array that gets fed into the GAN network.

sdk/opendp/whitenoise/synthesizers/pytorch/nn/dpctgan.py
```diff
@@ -0,0 +1,331 @@
...
+                c1 = torch.from_numpy(c1).to(self.device)
+                fakez = torch.cat([fakez, c1], dim=1)
+
+            fake = self.generator(fakez)
+            fakeact = self._apply_activate(fake)
+            data.append(fakeact.detach().cpu().numpy())
+
+        data = np.concatenate(data, axis=0)
+        data = data[:n]
+        return self.transformer.inverse_transform(data, None)
```
anujdesai510, line 331: return data. inverse_transform(data, None) should get handled by PytorchDPSynthesizer.generate(). The idea is to allow a user to choose any preprocessor and GAN they want without forcing them to use either.

sdk/opendp/whitenoise/synthesizers/pytorch/nn/__init__.py
AprilXiaoyanLiu: Hi Anuj, the reason that we need categoircal_columns and oridinal_columns here for CTGAN is because ctgan needs information from transformer for sampling and conditional generator, "self.transformer = DataTransformer()
        self.transformer.fit(train_data, categorical_columns)
        train_data = self.transformer.transform(train_data)

        data_sampler = Sampler(train_data, self.transformer.output_info)" 

so transforrmed data within train step instead of transform first and then pass prepocessed data in. Or we can pass preprocessor in instead of pass categorical column in

> We shouldn't need categorical_columns and ordinal_columns as arguments here. They're only used to indicate how preprocessing gets handled but the GAN classes should simply take a matrix that has already been transformed as input to train()

sdk/opendp/whitenoise/synthesizers/pytorch/nn/dpctgan.py
AprilXiaoyanLiu: > You should be able to remove this and call PytorchDPSynthesizer(DataTransformer(), DPCTGAN()), where the output of DataTransformer is the numpy array that gets fed into the GAN network.

similar reason as before regarding why do preprocessing data/transforming data in train step

eedeleon:
Forcing the Squash Merge since vsts seems to be missing the connection although the tests related to the PR all passed.

PR #259, opened 2020-08-11
Build and test Interface for OpenDP Core and support for multiple keys

ankit-oss:
WhiteNoise Benchmarking Multikey Support for testing tau-thresholding, ranking functions, GROUP BY queries
--------------------------------------
1. Creating PrivacyAlgorithm interface for OpenDP Core library and testing it by passing through the stochastic evaluator and benchmarking interfaces

2. When evaluation and benchmarking happen, it is not necessary, it is for a singleton result. There could be a GROUP BY or other partitions carrying multi-key results. So for evaluating such cases, we included the capability to iterate through all the keys provided by the user and evaluate the results for each one separately. 

3. Create actual_release function instead of actual=True parameter in PrivacyAlgorithm interface

4. Supporting multiple algorithms behind a single PrivacyAlgorithm interface to be benchmarked in one call

5. Adding bias test

6. Creating notebook showcasing benchmarking metrics vector and plots of algorithms for aggregations like SUM, MEAN in OpenDP Core library

7. Made changes to switch from whitenoise to smartnoise

Note: The tests using OpenDP Core repository are skipped by default as they are mainly for illustrative purposes and not needed to be part of OpenDP System build. 

**Result**

(py38) ankit@ANKIT-DELL:/mnt/c/Users/ankitsri/source/code/privacytoolsproject/whitenoise-system/tests/sdk/evaluation$ pytest -k "test_interface" -s
============== test session starts ==================================================
platform linux -- Python 3.8.5, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
rootdir: /mnt/c/Users/ankitsri/source/code/privacytoolsproject/whitenoise-system/tests, configfile: pytest.ini
collected 18 items / 11 deselected / 7 selected

test_core.py ..
test_interface.py ....
test_sql.py .

================ 7 passed, 11 deselected in 9.43s ===========================================


joshua-oss (approved):

ankit-oss:
Thank you Josh!

PR #267, opened 2020-08-28
dataverse_uri_loader function added

mikephelan:


sdk/opendp/whitenoise/data/dataverse_adapter.py
```diff
@@ -28,6 +29,27 @@ def dataverse_loader(host, doi=None, token=None):
...
+        scheme = uri.scheme
+        host = uri.netloc
+        path = uri.path
+        query = uri.query
+    except Exception as e:
+        print(str(e))
+
+    request = urllib.requests.get(scheme + '//' + host + path + '?' + query)
+    if(token):
+        request.add_header('X-Dataverse-Key', token)
```
eedeleon, line 44: I think we should pass the token in the query and extract it to pass as the header here

eedeleon: That would allow us to have 1 param for the uri, although, as mentioned before, this might not be the right way longer term if we can give the compute access

eedeleon:
A separate PR added the functionality

PR #272, opened 2020-09-08
April/updateopacus

AprilXiaoyanLiu:
- updated opacus after openning a new PR

### DIFF
--- a/sdk/opendp/whitenoise/synthesizers/pytorch/nn/__init__.py
+++ b/sdk/opendp/whitenoise/synthesizers/pytorch/nn/__init__.py
@@ -119,14 +119,14 @@ class DPGAN:
                 noise = torch.randn(self.batch_size, self.latent_dim, 1, 1, device=self.device)
                 noise = noise.view(-1, self.latent_dim)
                 fake_data = self.generator(noise)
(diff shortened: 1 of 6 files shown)


sdk/opendp/whitenoise/synthesizers/pytorch/nn/__init__.py
```diff
@@ -8,7 +8,7 @@
 import torch.optim as optim
 from torch.utils.data import DataLoader, TensorDataset
 
-from torchdp import PrivacyEngine, utils, autograd_grad_sample
+from opacus import PrivacyEngine, utils, autograd_grad_sample
```
eedeleon, line 11: we should update the requirements.txt that need opacus to replace the pytorch-dp dep with opacus. 

I suspect that will fix the builds

eedeleon: I added the change with github ui :)

lurosenb (approved):
Tests added, fixed the build conflicts - everything looks good.

Had to comment out a test from test_query (test_group_by_noisy_order), may have to do with updating the python version. @eedeleon @joshua-oss maybe touch base about this. Thanks!

PR #274, opened 2020-09-10
Add new dsets, update eval pipeline pre final runs 

lurosenb:


sdk/opendp/whitenoise/synthesizers/pytorch/nn/dpctgan.py
```diff
@@ -105,11 +105,9 @@ def __init__(self,
                  target_delta=None,
                  sigma = 5,
                  max_per_sample_grad_norm=1.0,
-                 epsilon = 3,
+                 epsilon = 1,
```
eedeleon, line 108: we set this? It is not configurable?

eedeleon (approved):
Looks good!

PR #276, opened 2020-09-13
Eedeleon/setup mlflow benchmarking

eedeleon:
Update eval pipeline to run as reproducible projects


lurosenb (approved):
The pipeline now essentially requires azureml-mlflow integration, but the runs are way faster if leveraged correctly. TODO: Add local option that swaps back to the joblib Parallel option (still the fastest way to run on a local machine). TODO: Add documentation for running through mlflow/azureml. TODO: Update plotting/visualization code.

PR #281, opened 2020-09-18
remove cross entropy in generator loss

AprilXiaoyanLiu:
to avoid potential privacy leak
I don't think patectgan needs it, but made change just in case


lurosenb (approved):
Ok yes - let's add this for now to be safe, and reinstate once verified.

PR #284, opened 2020-09-23
Wrong Classname for PytorchSynthesizer

Tecnarca:
The test fails at line 31 because of the wrong classname.
Other classes implement the correct one.


eedeleon (approved):
Nice find! Thanks for the fix

PR #286, opened 2020-09-29
Added Argument parser for DPSDGYM

Tecnarca:
Modified the command line interface to take as input only specific datasets, epsilons and metrics to run.
Moved the default array for metrics and epsilons in config.py
Updated README.md to match current input system.

Possible extension: allow one to select models to be run (from the ones in config.py) and let one select the mlflow run name (currently defaults to "test").


tests/dpsdgym/README.md
```diff
@@ -56,17 +56,17 @@ Make sure you have the following installed in a conda env, before running the ev
...
 Once the configuration/datasets are set up, the entire evaluation pipeline, with all metrics, can be run through the eval script
 ```
-python main.py all
+python main.py
 ```
-Flags can be used to specify which metrics to run in the evaluation. You could equivalently run
+Flags can be specified to specify which metrics, epsilon and datasets should be used to run the evaluation:
 ```
-python main.py wasserstein pmse ml_eval sra
+python main.py -d bank adult -e 0.01 0.1 1 -m pmse ml_eval wasserstein
```
eedeleon, line 63: Part of the reason we avoided this approach is that mlflow did not properly parse the arguments for the remote scenario.

Can you run these changes with mlflow run "tests/dpsdgym" -P {param}={val} to verify the changes are compatible?

eedeleon: I am going to assume this comment is not a blocker since those configuration points were not there to begin with

tests/dpsdgym/main.py
```diff
@@ -3,6 +3,9 @@
 import time
 import mlflow
 import json
+import argparse
+import textwrap
+
```
eedeleon, line 8: We will need to update the MLproject file in this directory to reflect the changes to main.py's CLI interface

Tecnarca:
I have corrected the argument parsing for MLFlow by parsing further the arguments when multiple values come in a single string as MLFlow was doing.
The command: 
`mlflow run "tests/dpsdgym" -P dataset="bank adult mushroom" -P epsilons="1 2" -P eval_metrics="mse ml_eval"`
Now runs the project correctly. I've updated the MLProject file accordingly to reflect this change.

I forgot to mention why this change is needed: it allows to specify epsilons prerun and this is is going to be important when we will introduce a checkpointing system in the future (otherwise we would not be able to run automatically different experiments, we would have to modify conf.py manually). It also allows to run multiple metrics and datasets in the same run, when needed.

tests/dpsdgym/benchmark.py
```diff
@@ -0,0 +1,41 @@
...
+	    if 'wasserstein' in metric:
+	        run_wasserstein(data_dicts, 100, run_name)
+
+	    if 'pmse' in metric:
+	        run_pMSE(data_dicts, run_name)
+	    
+	    if 'ml_eval' in metric:
+	        results = run_ml_eval(data_dicts, epsilons, run_name)
+	        
+	        if 'sra' in metric:
```
eedeleon, line 35: Is this supposed to be tabbed in?

tests/dpsdgym/main.py
```diff
@@ -1,52 +1,34 @@
...
+    parser.add_argument('-e', '--epsilon', nargs="+", default=conf.EPSILONS, help="Epsilons values for which the models will be evaluated")
+    parser.add_argument('-m', '--metric', nargs="+", default=conf.KNOWN_METRICS, help="Differential privacy metrics for which the models will be evaluated")
+    parser.add_argument('-r', '--run', default="dpbench_run", help="Run name for mlflow")
+    parser.add_argument('-p', '--path', default="results/", help="Path of the folder to save the results")
 
-from synthesis import run_all_synthesizers
-from evaluate import run_ml_eval, run_wasserstein, run_pMSE
+    args = parser.parse_args()
+    if isinstance(args.dataset[0], str) and len(args.dataset)==1:
+        args.dataset = args.dataset[0].split()
```
eedeleon, line 20: This is ok for now but comma separated strings will probably be better. Dataset names are more likely to have a space than a comma

tests/dpsdgym/main.py
```diff
@@ -1,52 +1,34 @@
...
 
-from warnings import simplefilter
-simplefilter(action='ignore', category=FutureWarning)
+    if isinstance(args.epsilon[0], str) and len(args.epsilon)==1:
+        args.epsilon = args.epsilon[0].split()
```
eedeleon, line 23: the space comment is less of an issue here :)

tests/dpsdgym/main.py
```diff
@@ -1,52 +1,34 @@
...
-    if 'pmse' in flags:
-        run_pMSE(data_dicts, run_name)
-    if 'ml_eval' in flags:
-        results = run_ml_eval(data_dicts, epsilons, run_name)
-        print(results)
-    with open('artifact.json', 'w') as f:
-        json.dump(results, f)
-    mlflow.log_artifact("artifact.json", "results.json")
+    if isinstance(args.metric[0], str) and len(args.metric)==1:
+        args.metric = args.metric[0].split()
```
eedeleon, line 26: This can probably be a util function args.metric = split_if_needed(args.metric)

tests/dpsdgym/synthesis.py
```diff
@@ -39,13 +39,13 @@ def run_synthesis(synthesis_args):
...
                          "epsilon": str(e),
-                         "dataset": str(datasets),
+                         "dataset": str(d),
                          "duration_seconds": str(end_time - start_time)})
         print(datasets[d]["name"] + ' finished. Epsilon: ' + str(e))
         datasets[d][n][str(e)] = sampled
     return (n, d, str(e), sampled)
 
-def run_all_synthesizers(datasets, epsilons):
+def run_all_synthesizers(datasets, epsilons, save_models_path, run_name):
```
eedeleon, line 48: Can the new parameters be optional with defaults?

eedeleon (approved):
These are great additions. Thank you so much for taking the time to add so many configuration and extension point possibilities to the Eval.

I think the only part that might leave the eval outputs hard to track is sra use + ml_eval results being modified. This change is ok for now though hence the approve since the comments are mostly around style

Issue #278, opened 2020-09-16
PATECTGAN has loss function wasserstein, but does not support 


lurosenb:
Switching to the wasserstein loss function in PATECTGAN breaks the synthesizer.
Either need to support wasserstein, or remove the option

Tecnarca:
I know that removing the option would be easier, but i would like to discuss the possibility of implementing it. 
From the original PATEGAN paper, i get that a wasserstein loss can't easily be used on the teachers (it would require to compute and prove a new bound for the privacy cost) but it can in principle be used on the student. I do not know the implications of doing something like this, but from my understanding it does not compromise the privacy guarantee.

I also do not know if someone is working on this, but i would like to help in extending this model.

eedeleon:
This is great to hear! @Tecnarca  we do not have someone working on this. The team would be happy to review a proposal and help guide you through contributing to the repo if you are interested.

Thanks for reaching out. We have not formalized the process of proposals, so for now a linked github issue with a proposal(if you would like to collaborate on it we can also arrange for that) can get the ball rolling.

AprilXiaoyanLiu:
This is great to hear +1!  @Tecnarca  Look forward to reviewing your proposal and the collaboration!

Tecnarca:
I have proposed the possible extension [here](https://github.com/opendifferentialprivacy/whitenoise-system/pull/288). I hope it is enough to provide a base to further improve from.

eedeleon:
Resolved in https://github.com/opendifferentialprivacy/smartnoise-sdk/pull/288

PR #288, opened 2020-10-06
PATECTGAN Model Extension (Wasserstein Loss)

Tecnarca:
Follow-up on issue #278 

This is the implementation that extends the PATECTGAN model.

Improved base implementation:
- The Generator of the CTGAN model requires to penalize its loss to force it to generate the correct vector given the conditional vector (page 5 under "conditional generator" of the [CTGAN Paper](https://arxiv.org/pdf/1907.00503.pdf)). For some reason this was not implemented here, I have added the call to the function from the CTGAN library (line 308).
- The generator was being trained in the same loop as the student model. Since this is not the case in the original CTGAN implementation and it is not the case in algorithm 1 from the PATEGAN paper, I imagine this is an indentation error

The class now implements a WGAN:
- The teachers could immediatly use this loss to learn from true data
- In the student, to compute the wasserstein loss we would need to sample from the true data, but since this would break the privacy requirement, we can only use the fake samples labeled as reals by the teachers. The idea is to take the fake samples from the generator and since they look real to the teachers, the student should agree with this evaluation.
- The generator uses the same loss of the student, but only the first term, as in a normal WGAN (plus the cross entropy mentioned above).

The class also implements the [DRAGAN](https://arxiv.org/pdf/1701.07875.pdf) regularization:
- A WGAN architecture would in principle need the Gradient Penalty added to the discriminator. In this particular framework, we cannot implement that regularization: the gradient penalty regularization is based on pairs of true and fake (generated) samples. This is a problem for the student, since its wasserstein implementation has a number of samples that look real based on what the teachers say. This means that we do not get an equal number of true and fake samples in general
- To solve the above issue, one could oversample from the generator until we reach the same number of samples for both the true and fake labels given by the teachers, than discard the rest of the generated samples. However, this is not possible, since to sample from the generator we are also sampling a conditional vector from the real distribution of the data. If we discard some samples after the generation, we are in fact altering the distribution of labels that we are learning from.
- Both the above issues are solved by the DRAGAN regularization. See the DRAGAN paper for a detailed explanation and theorical derivation. The idea of DRAGAN is to regularize (by forcing the gradient to be 1) only in a contour of the true data points (achieved by adding a noise to the real data point and computing the gradient there, as in WGAN-GP).
- The dragan regularization allows us to take the fake samples labeled as true by the teachers (as in the WGAN above) and, assuming they look  "real enough", compute the penalty to the loss based on them. This kind of regularization can be used on the teachers (that use truly real data) and by the student (as i just described) to enhance the capabilites of the network to learn.

In all these modifications, I did not touch any of the requirements of the PATEGAN framework, so this implementation is still completely differentialy private (the only part where we see real data is when the teachers perform a vote, and epsilon is computed based on the number of these votes).

Note that I have made only small test of this architecture (with both WGAN and DRAGAN active), but comparing the correlations of the synthetic and true data i've saw that the architecture with WGAN and DRAGAN active performs better (more closely reproduces the correlations) than the implementation with only the binary cross entropy loss. 

Further steps for this implementation would be to properly test it (and i will do so of both the privacy and the performance using DPSDGYM, comparing it also with CTGAN, PATEGAN and WGAN). Another improvement could be to modify the pate mechanism to implement a gaussian noise (along with other improvements) as described [in this paper](https://arxiv.org/pdf/1802.08908.pdf).

Let me know if you find any error or if you have any suggestion. I would be happy to answer and find a way to improve this model further.


Tecnarca:
If it helps advancing faster, I am willing to schedule a meeting to talk about this proposed architecture and how to evlolve it further.

AprilXiaoyanLiu:
> If it helps advancing faster, I am willing to schedule a meeting (my timezone is CET+1) to talk about this proposed architecture and how to evolve it further.

Hi Tecnarca, 

Thank you for providing such detailed and clean code for us to review. I took a quick glance it, it looks good to me. I will test this out sometime next week and get back to you (I've been a bit busy this week). I will let you know if I have some questions. Thanks again!

Tecnarca:
Since last week, I found an indentation error (?) for the PATECTGAN model.
I've added it in the first section of the description above.

eedeleon:
I am going to add a smoke test so we can check in your progress before we make a large, conflicting PR

eedeleon:
@Tecnarca  it would be great for us to schedule a time to meet sometime next week. I will be checking in this PR today, but hopefully we can make it easier for you to contribute further enhancements and evaluations in the future if you are interested.

eedeleon:
Alright, during testing we found 

 RuntimeError: all elements of input should be between 0 and 1

April and I are looking into these to see if we need to normalize the data to be within that range or not.

eedeleon:
We could not resolve the conflict unfortunately. The conflict was a rename of the directory and a replacement of the prior name whitenoise with the new project name smartnoise

Tecnarca:
> We could not resolve the conflict unfortunately. The conflict was a rename of the directory and a replacement of the prior name whitenoise with the new project name smartnoise

I should have resolved the conflict for merging the two repos now, I hope that also fixes the problem with the Azure Pipeline. Let me know wether this is the case or not.

> @Tecnarca it would be great for us to schedule a time to meet virtually sometime next week. I will be checking in this PR today, but hopefully we can make it easier for you to contribute further enhancements and evaluations in the future if you are interested.

I would love to contribute further. Let's meet! I am usually available monday/tuesday/wednesday/friday anytime between 9:00 am and 6:00 pm (CET), but we can easily agree on other slots if necessary. If you want to schedule a meeting, send me an invite to mail.tecnarca@gmail.com and we will take it from there.

eedeleon:
@Tecnarca  I will be reaching out over email. Exciting to hear. Looking forward to the meeting

Tecnarca:
@eedeleon I have made another commit, where I have added a test for the Wasserstein Loss w/ Dragan regularization and for dragan regularization alone. Also, in the regularization parameter, I now override the regularization with "Dragan" when a wasserstein loss is used (this prevents the neural network to be trained incorrectly using a wasserstein loss, that requires a regularization). The tests that i've run have shown some problems between CPU and GPU tensors, which I have procedeed to fix. I never got the error of the weights being negative, I think it could be because of the Wasserstein loss. 

Please let me know if the error still occours.

eedeleon:
This is a great contribution. I will be working on getting the pytorch evaluators running at the gates soon. Thanks for adding the test case!

We will run our evaluation suite with the new method soon to see how it compares with our previously compared methods.

eedeleon (approved):
Great contribution!

eedeleon:
The smartnoise build passed.

PR #291, opened 2020-10-22
Rename to smartnoise

eedeleon:


eedeleon:
Tests passed and the files were string replaced whitenoise with smartnoise and a renamed directory

PR #293, opened 2020-10-29
Whitepaper for SQL accuracy and budget

joshua-oss:
Adding paper with details on accuracy bounds and budget for SQL layer


eedeleon (approved):
This is super useful

Issue #206, opened 2020-05-21
Consider hiding/removing some override flags


joshua-oss:
Override flags are available for cases when the underlying data is known to be safe, e.g. if the curator knows the underlying database enforces column values or uniqueness.  These are currently exposed though the API.  Consider removing these from the API and making them curator-specified via metadata, or otherwise mitigating possibility of mistaken use.

joshua-oss:
PrivateReader is designed to be used in a trusted context.  For example, the non-private Reader object is available directly on private.reader, and contains secure connection information as well as direct access to private data.  In untrusted scenarios, PrivateReader would be isolated behind a hardened trust boundary such as REST over TLS, with the analyst controlling only the subset of queries supplied to the pre-configured object.

However, analysts in the notebook context may not understand the implications of some options, and could inadvertently release unsafe reports.  Current proposal is to make some changes to make the distinction between curator and analyst more explicit:

1. Remove ability to pass in MetadataCollection to PrivateReader.  Metadata will only be available via persisted file.  Curator will be able to manipulate MetadataCollection programmatically, but must persist to disk before analyst can use.
2.  Curator can still indicate optimizations, such as reservoir_sample and clamp_counts, if the source storage engine enforces the necessary constraints.  These will be passed via metadata and used by the PrivateReader.
3.  Analyst will not be able to pass these options through the API.  Analyst can still pretend to be the curator and make fake metadata, but this is always a possibility when analyst is trusted.

PR #299, opened 2020-11-03
Safe settings

joshua-oss:
Disable analyst setting of override flags, per #206.  All overrides must be set in YAML.

Added detailed documentation for Metadata file format at: sdk/Metadata.yaml

This PR introduces multiple breaking changes:
* If analyst was calling setters on private.options, the code will now fail, and tell the analyst to change settings in metadata
* Calling pattern for instantiating PrivateReader and PandasReader now passes the metadata as a the second parameter, so that this can be optional in the future.  The "df" or "reader" param is the only required parameter.
* CollectionMetadata instantiation puts the rowcount after the columns.  This is because rowcount is now optional.  This should not impact users, since CollectionMetadata is no longer intended to be constructed directly.


sdk/Metadata.md
```diff
@@ -0,0 +1,171 @@
...
+* `clamp_columns`: Boolean.  Default is `True`.  By default, the system clamps all input data to ensure that it falls within the `lower` and `upper` bounds specified for that column.  If the data curator can be certain that the data never fall outside the specified ranges, this step can be disabled.
+* `use_dpsu`: Boolean.  Default is `False`.  Tells the system to use Differential Private Set Union for censoring of rare dimensions.  Does not impact privacy.
+
+## Column Options
+
+* `type`: Required. This type attribute indicates the simple type for all values in the column. Type may be one of “int”, “float”, “string”, “boolean”, or “date”. The “date” type includes date or time types.  If type is set to "unknown", the column will be ignored by the system.
+* `private_key`: Boolean.  Default is `False`.  indicates that this column is the private identifier (e.g. “UserID”, “Household”).  This column is optional.  Only columns which have private_id set to ‘true’ are treated as individuals subject to privacy protection.
```
eedeleon, line 58: nit: casing on Indicates

sdk/opendp/smartnoise/metadata/collection.py
```diff
@@ -214,13 +221,16 @@ def load_table(self, schema, table, t):
         row_privacy = bool(t["row_privacy"]) if "row_privacy" in t else False
         max_ids = int(t["max_ids"]) if "max_ids" in t else 1
         sample_max_ids = bool(t["sample_max_ids"]) if "sample_max_ids" in t else None
+        use_dpsu = bool(t["use_dpsu"]) if "use_dpsu" in t else None
+        clamp_counts = bool(t["clamp_counts"]) if "clamp_counts" in t else None
```
eedeleon, line 225: Why is clamp_columns default a bool and clamp_counts default None?

eedeleon: Same for dpsu

sdk/opendp/smartnoise/sql/private_reader.py
```diff
@@ -32,9 +32,17 @@ def __init__(self, metadata, reader, epsilon_per_column=1.0, delta=10E-16,
             :param options: A PrivateReaderOptions with flags that change the behavior of the privacy
                 engine.
         """
-        self.options = options if options is not None else PrivateReaderOptions()
-        self.metadata = metadata
-        self.reader = reader
+        if isinstance(reader, Reader):
```
eedeleon, line 35: I think for the first release of this change we can log a warning if the order is flipped.

if type(metadata, Reader) and type(reader, CollectionMetadata):
     log a warning
      tmp = reader
      reader = metadata
      metadata = tmp

This could be added before the current code for now. And removed later

sdk/opendp/smartnoise/sql/private_reader.py
```diff
@@ -32,9 +32,17 @@ def __init__(self, metadata, reader, epsilon_per_column=1.0, delta=10E-16,
...
+            self.reader = reader
+        else:
+            raise ValueError("Parameter reader must be of type Reader")
+
+        # using string here, because we don't want to import .metadata due to circular reference
+        if "metadata.collection.CollectionMetadata" in str(type(metadata)):
```
eedeleon, line 41: isinstance(metadata, metadata.collection.CollectionMetadata)

tests/sdk/test_local_analyst.py
```diff
@@ -10,15 +10,16 @@ def test_sklearn_query():
...
+   iris = Table("dbo", "iris", [
       Float("sepal length (cm)", 4, 8),
       Float("sepal width (cm)", 2, 5),
       Float("petal length (cm)", 1, 7),
       Float("petal width (cm)", 0, 3)
-   ])
+   ], 150)
    schema = CollectionMetadata([iris], "csv")
 
-   reader = PandasReader(schema, sklearn_df)
```
eedeleon, line 21: Can we duplicate this test with the older paradigm to verify the back compat for 1 release.

sdk/opendp/smartnoise/metadata/collection.py
joshua-oss: The code defaults to None for those params if they aren't present, since they are optional in metadata, and we want the metadata to round-trip back to disk without adding properties that weren't set originally.

sdk/opendp/smartnoise/sql/private_reader.py
joshua-oss: isinstance doesn't work here, because you would need to import it, and that creates a circular reference.

sdk/opendp/smartnoise/metadata/collection.py
eedeleon: makes sense

eedeleon (approved):
This API seems more intuitive.

Thanks for also removing the overrides to protect against privacy reducing mistakes!

PR #304, opened 2020-11-09
Use core

joshua-oss:


sdk/opendp/smartnoise/evaluation/_aggregation.py
```diff
@@ -79,27 +78,23 @@ def dp_var(self, df, colname):
...
+        Returns repeatedly applied noise adding mechanisms to sum query.
         Sensitivity is set as absolute difference between max and min values
         within the column
         """
         exact_sum = df[colname].sum()
         M = float(abs(max(df[colname]) - min(df[colname])))
-        mech = Laplace(self.epsilon, sensitivity = M)
         if(self.mechanism == "Gaussian"):
-            mech = Gaussian(self.epsilon)
+            mech = Gaussian(self.epsilon, 10E-5, M)
```
eedeleon, line 97: We should remove the configurability of sel.mechanism. Otherwise passing Laplace would end up in an undefined error for mech which would be hard to debug

sdk/opendp/smartnoise/metadata/collection.py
```diff
@@ -220,17 +222,18 @@ def load_table(self, schema, table, t):
         rows_exact = int(t["rows_exact"]) if "rows_exact" in t else None
         row_privacy = bool(t["row_privacy"]) if "row_privacy" in t else False
         max_ids = int(t["max_ids"]) if "max_ids" in t else 1
-        sample_max_ids = bool(t["sample_max_ids"]) if "sample_max_ids" in t else None
-        use_dpsu = bool(t["use_dpsu"]) if "use_dpsu" in t else None
-        clamp_counts = bool(t["clamp_counts"]) if "clamp_counts" in t else None
+        sample_max_ids = bool(t["sample_max_ids"]) if "sample_max_ids" in t else True
+        use_dpsu = bool(t["use_dpsu"]) if "use_dpsu" in t else False
```
eedeleon, line 226: this looks like a bad merge?

sdk/opendp/smartnoise/sql/private_reader.py
```diff
@@ -10,7 +10,7 @@
 from opendp.smartnoise._ast.expressions import sql as ast
 from opendp.smartnoise.reader import Reader
 
-from opendp.smartnoise.mechanisms.gaussian import Gaussian
+from opendp.smartnoise._mechanisms.gaussian import Gaussian
```
eedeleon, line 13: it might be better to move _mechanisms here

opendp/smartnoise/sql/_mechanisms.py

And to use the relative import ._mechanisms

No other part of the code base uses it and we don't expect further use right?

sdk/opendp/smartnoise/sql/private_reader.py
```diff
@@ -202,7 +203,7 @@ def _execute_ast(self, query, cache_exact=False):
             kc_pos = kcc_pos.pop()
 
         # make a list of mechanisms in column order
-        mechs = [Gaussian(self.epsilon_per_column, self.delta, s, max_contrib, self.interval_widths) if s is not None else None for s in sens]
```
eedeleon, line 205: we should keep test_gauss?

sdk/opendp/smartnoise/metadata/collection.py
joshua-oss: No, it's all there if you expand.  I had to add censor_dims, since that was used in the Spark Private Synopsis, and was previously only available in PrivateReaderOptions, not metadata

joshua-oss: It's good; you can see all the properties if you expand

sdk/opendp/smartnoise/evaluation/_aggregation.py
joshua-oss: Good call; I nuked it all

sdk/opendp/smartnoise/sql/private_reader.py
joshua-oss: test_gauss was testing the proper noise distribution when sensitivity changed.  That's handled in core library tests now

sdk/opendp/smartnoise/sql/private_reader.py
joshua-oss: That's correct; I've moved it

sdk/opendp/smartnoise/metadata/collection.py
eedeleon: makes sense, the change from None to True / False is what confused me

sdk/opendp/smartnoise/sql/dpsu.py
```diff
@@ -6,7 +6,7 @@
 
 from opendp.smartnoise.sql.parse import QueryParser
 from opendp.smartnoise._ast.ast import Table
-from opendp.smartnoise.mechanisms.rand import laplace
+from opendp.smartnoise.sql._mechanisms.rand import laplace
```
eedeleon, line 9: this can also use the ._mechanisms import but is mostly a nit

eedeleon (approved):
This will resolve a github issues as well.

I am really happy to see this change happen already. Looking forward to see what improvements in utility and future work this contribution will bring.

Thanks!

sdk/opendp/smartnoise/sql/dpsu.py
joshua-oss: Fixed

Issue #287, opened 2020-10-01
HAVING clause is ignored by PrivateReader


gregaw:
Given table `A: (id int, product text, quantity int)`, when you run the following query with a PrivateReader:

```
select product, SUM(quantity) as sm
from A 
group by product
having IDONTEXIST > 1
```
then it returns results as if ignoring the incorrect HAVING statement.

When the same sql runs via a PostgresReader directly, then the appropriate exception is thrown:
`psycopg2.errors.UndefinedColumn: column "idontexist" does not exist`

eedeleon:
@joshua-oss  looks like our rewriter might be dropping having statements rather than failing early or rewriting them properly.

Thanks for letting us know @gregaw , apologies for the delayed response, we will look into this and report back on what can be done within a week

joshua-oss:
Oh, right, we can either fail early and tell people to manually filter the aggregates that are returned, or else just go ahead and implement it.  Because the HAVING clause needs to be applied in post-processing, after the differentially private aggregates are computed, it's not difficult.  The main work is in the validator.

eedeleon:
We have checked in HAVING support. We will close this issue when it reaches pypi

gregaw:
thanks @eedeleon. Appreciated 👍

joshua-oss:
In PyPi

Issue #30, opened 2019-12-02
Implement TOP K


joshua-oss:
SQL syntax to allow TOP K.  May be implemented in post-processing and needn’t use a specific heavy hitters algorithm, but API should support plugging in different heavy hitters or top K algorithms

PR #306, opened 2020-11-11
Support for HAVING, TOP K, and LIMIT

joshua-oss:
These are all performed in post processing.

Addresses #287 and #30.


sdk/opendp/smartnoise/_ast/expressions/logical.py
```diff
@@ -29,7 +32,56 @@ def children(self):
...
+                    try:
+                        l = int(l)
+                    except:
+                        l = float(l)
+                elif isinstance(r, float):
+                    l = float(l)
+                elif isinstance(r, datetime):
+                    l = datetime.fromisoformat(l)
+                elif isinstance(r, date):
+                    l = date.fromisoformat(l)
```
eedeleon, line 49: The nested ifs should be a function.

if isinstance(l, str):
   l = resolve_string_type(l, r)
elif isinstance(r, str):
   r = resolve_str_type(r, l)

eedeleon (approved):
This feature has been asked for quite a bit.

Thanks for the contributions!

Do we need to update any md for how the code is working?

joshua-oss (commented):
> This feature has been asked for quite a bit.
> 
> Thanks for the contributions!
> 
> Do we need to update any md for how the code is working?

I actually took the liberty of updating the design paper in the /papers/ folder as part of the safe-settings PR.  So the implementation now matches the paper.

sdk/opendp/smartnoise/_ast/expressions/logical.py
joshua-oss: Good call; I have updated

PR #307, opened 2020-11-14
Support unit tests against supported Readers

joshua-oss:


eedeleon (approved):
Nice. This will make it easier for us to add these to a build too!

PR #313, opened 2020-11-21
Typed to pandas

joshua-oss:
Eliminated TypedRowset and use Pandas


tests/sdk/evaluation/dp_singleton_query.py
```diff
@@ -30,12 +29,11 @@ def release(self, dataset: object) -> Report:
         """
         private_reader = PrivateReader(dataset[1], dataset[0], self.privacy_params.epsilon)
         query_ast = private_reader.parse_query_string(self.algorithm)
-        srs_orig = private_reader.reader.execute_ast_typed(query_ast)
```
eedeleon, line 33: Why are internals needed here?

sdk/opendp/smartnoise/reader/base.py
```diff
@@ -1,55 +1,27 @@
...
-            query_string = str(query)
-        return self.execute(query_string)
+            return pd.DataFrame(rows[1:], columns=rows[0])
 
-    def execute_ast_typed(self, query):
-        syms = query.all_symbols()
-        types = [s[1].type() for s in syms]
+    def execute_df(self, query):
+        if not isinstance(query, str):
+            raise ValueError("Please pass a string to this function.  You can use execute_ast to execute ASTs")
```
eedeleon, line 25: We should remove the second half of the error message now that ast usage is iternal

eedeleon (approved):
PR looks good.

The only part that would be worth looking into later is what is internal and what is not.

ast usage in the evaluation seems off but is not a blocker. Ideally secondary features based off sql would be usable without internal functions. If that is not the case, we may need to promote a subset of the ast logic to public

tests/sdk/evaluation/dp_singleton_query.py
joshua-oss: I don't expect analysts to ever need to pass in a parsed AST.  Only the evaluator or developers of custom readers will need these.  This simplifies the Reader public interface to have just execute() and execute_df()

sdk/opendp/smartnoise/reader/base.py
joshua-oss: Good catch; fixed

tests/sdk/evaluation/dp_singleton_query.py
joshua-oss: The internal methods are used on the evaluator for two reasons:  first, to avoid the perf hit of parsing the query and resolving symbols 10,000 times for every run, and second, to allow the cache_exact field to store the exact query results internally (for the accuracy checker).

The first use-case might be relevant to analysts, if there is a scenario that involves calling the same query multiple times per second.

The second use-case is peculiar to the evaluator, and is really not a great design.  We have discussed with James and Michael about switching to a convention where any API that receives infinity for epsilon will, by convention, treat that as a request for a non-private answer.   If we can get a decision on some convention like that, we could eliminate the other use case that requires ASTs.

PR #315, opened 2020-11-23
Early quail

lurosenb:


sdk/opendp/smartnoise/synthesizers/quail.py
```diff
@@ -0,0 +1,108 @@
...
+    and use synthetic data from the DP synthesizer with
+    the DP classifier to create artificial labels. Produces
+    complete synthetic data.
+
+    More information here: 
+    Differentially Private Synthetic Data: Applied Evaluations and Enhancements
+    https://arxiv.org/abs/2011.05537
+    """
+    
+    def __init__(self, epsilon, dp_synthesizer, synth_args, dp_classifier, class_args, target, test_size=0.2, seed=42, eps_split=0.9):
```
eedeleon, line 24: It might be a simpler API to have

(epsilon,  lambda epsilon: returns synthesizer, lambda epsilon: return classifier, target, seed, eps_split)

I do not think we need the train test split done internally within QUAIL

eedeleon: And func that returns the object is more flexible on inputs as well as epsilon handling.

For example, how would key word arguments work with synth_args?

eedeleon: The public API should include
`

@property 
def classifier/model

@property 
def synthesizer`

Also a test please :)

I am so excited to ship this feature. Great idea and thanks for the contribution!

The run time to utility tradeoff is one of the more exciting aspects of the approach

eedeleon (approved):
Nice! There are some minor param name improvements we might do later but this looks good enough to put in users hands so they can inform the param names!

This is awesome! Looking forward to the interesting use cases QUAIL can help with. Especially the training time reduction!

Issue #317, opened 2020-11-30
Not specifying column metadata bounds leaves a column with no noise applied.


gregaw:
To reproduce 
- create a table with two columns and the following smartnoise metadata:
```
col:
  ? ''
  : table:
      duration:
        type: float
      id:
        private_id: true
        type: int
      max_ids: 1
      row_privacy: false
      rows: 1000
      sample_max_ids: true
engine: pandas
```

- run multiple queries summing and averaging the `duration` column

The problem:
- noise gets applied to the AVG queries, but doesn't to the SUM queries (see the attached image)

The reason:
- sensitivity for the `duration` column is set to `None` and disables the noise mechanism
- `AVG` works only because the noise for `COUNT` is hardcoded to `1`.

![image](https://user-images.githubusercontent.com/361951/100630889-653c6800-332b-11eb-9603-5e4e02ae1484.png)

joshua-oss:
Great catch, thanks for reporting this.  We had a bug where the unbounded sensitivity in AST wasn't propagating up to the resolved query.  Fixed in PR #318.

gregaw:
wow - that's a quick turn-around 👍  Thanks!

eedeleon:
The change is also on pypi now

PR #318, opened 2020-12-01
Checks for output with unbounded sensitivity

joshua-oss:
Fix for #317.  AST was not propagating unbounded sensitivity to resolved symbols.  

When a column has unbounded sensitivity, we now allow queries like `COUNT(col`) and `SELECT col GROUP BY col`, but prohibit queries like `SUM(col)` or `AVG(col)`.


eedeleon (approved):
Nice bug fix!

eedeleon (approved):
We should be releasing this change soon

PR #320, opened 2020-12-03
Adding two columns should add the sensitivities

### LINKED ISSUE #319: Sensitivity calculation of a sum expression is inconsistent
class ArithmeticExpression calculates the sensitivity of an expression x+y as max (x.sens, y.sens) and sensitivity of 2*x as 2*x.sens.

But then we get 2*x.sens=sens(2*x) <> sens(x+x)=max(x.sens, x.sens)=x.sens.

joshua-oss:
Fixes #319.  Sensitivity of two columns added is the sum of sensitivities


eedeleon (approved):
Makes sense.

Nice bug fix

Issue #321, opened 2020-12-07
Example throwing sqlite3.OperationalError


SrivastavaAnubhav:
I'm trying to run the basic example in the README from https://github.com/opendifferentialprivacy/smartnoise-sdk, and I am getting this error. I'm not sure what this means:

```
Traceback (most recent call last):
  File "/home/anubhav/.local/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1267, in _execute_context
    cursor, statement, context
  File "/home/anubhav/.local/lib/python3.6/site-packages/sqlalchemy/engine/default.py", line 596, in do_execute_no_params
    cursor.execute(statement)
sqlite3.OperationalError: near "(": syntax error

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/anubhav/.local/lib/python3.6/site-packages/pandasql/sqldf.py", line 61, in __call__
    result = read_sql(query, conn)
  File "/home/anubhav/.local/lib/python3.6/site-packages/pandas/io/sql.py", line 515, in read_sql
    chunksize=chunksize,
  File "/home/anubhav/.local/lib/python3.6/site-packages/pandas/io/sql.py", line 1295, in read_query
    result = self.execute(*args)
  File "/home/anubhav/.local/lib/python3.6/site-packages/pandas/io/sql.py", line 1162, in execute
    *args, **kwargs
  File "/home/anubhav/.local/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1003, in execute
    return self._execute_text(object_, multiparams, params)
  File "/home/anubhav/.local/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1178, in _execute_text
    parameters,
  File "/home/anubhav/.local/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1317, in _execute_context
    e, statement, parameters, cursor, context
  File "/home/anubhav/.local/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1511, in _handle_dbapi_exception
    sqlalchemy_exception, with_traceback=exc_info[2], from_=e
  File "/home/anubhav/.local/lib/python3.6/site-packages/sqlalchemy/util/compat.py", line 182, in raise_
    raise exception
  File "/home/anubhav/.local/lib/python3.6/site-packages/sqlalchemy/engine/base.py", line 1267, in _execute_context
    cursor, statement, context
[...]

PR #322, opened 2020-12-08
Add warning for older sqlite3

joshua-oss:
PandasReader warns user on instantiation, if the version of sqlite3 installed is older than 3.2.6 (which I believe applies to Python 3.6 and some older versions of non-conda 3.7).  Recent conda picks up 3.31.1.

Queries with `row_privacy` work on older sqlite3, but people will need to upgrade or use conda to get `private_key`.

Fix for #321


eedeleon (approved):
Nice contribution. This is likely due to the conda vs pip install issue as well. 

Not sure how best to resolve this altogether

PR #329, opened 2021-01-11
update opendp.io to opendp.org

raprasad:
Update smartnoise@opendp.io to smartnoise@opendp.org
(Note, both addresses work and forward to the support channel)


eedeleon (approved):
Thanks for the change!

PR #331, opened 2021-01-22
Fix typo that breaks formatting

anthager:
Replace single quote with a backtick to apply correct md formatting


eedeleon:
Nice catch!

eedeleon (approved):
Thank you!

PR #338, opened 2021-02-23
Improve base docs

eedeleon:
- use wraps where applicable
- improve base synthesizer documentation
- reorganize code to keep init at the top


joshua-oss (approved):
Nice!

PR #339, opened 2021-03-01
Fix warning import error.

bwuzhang:
I believe the previous `import warning` is an error. It causes `from opendp.smartnoise.synthesizers.quail import QUAILSynthesizer` to fail.


eedeleon (approved):
Thank you! Nice catch, not sure why the builds did not catch this

PR #340, opened 2021-03-11
Improve user experience for synthesizers

eedeleon:
- use epsilon more broadly
- expose imports at top level, nn, synthesizers
- split up objects into multiple files
- Move epsilon to the first param


sdk/opendp/smartnoise/sql/__init__.py
```diff
@@ -25,7 +25,7 @@
 ]
 
 
-def execute_private_query(reader, schema, budget, query):
+def execute_private_query(reader, schema, epsilon, query):
```
joshua-oss, line 28: I'm just now noticing this doesn't pass in delta, which will result in the default delta from PrivateReader being used.  This may be OK, but could be either too conservative or too relaxed, depending on how many rows are in the source data.  The delta term impacts how many records get dropped in the tau thresholding.

joshua-oss (approved):
These improvements look great, this will simplify calling pattern.

Issue #345, opened 2021-03-29
Visualizing the ast graph could make debugging easier


vserret:
When rewriting a parsed query for making it differentially private, some branches of the original AST are handwritten and this could be a source of bugs. 
The rewritten query should not change when parsed, i.e.`private_query == QueryParser().parse(str(private_query)))` must be `True`. 

**Visualizing a parsed query could make debugging easier by identifying at a glance the incorrect nodes of the AST.** 

Example:

The parsed `SELECT SUM(age) AS my_sum FROM pums.pums GROUP BY age` could be represented as: 

![parsed_query](https://user-images.githubusercontent.com/22152801/112838426-54afba80-909d-11eb-816b-ff278edb27ba.png)

If we break the grammar rules by doing `query.agg = 'GROUP BY age'`, the GROUP BY clause of the parsed query is now a `str` instead of an `Aggregate` object. 
By displaying `private_query` and the `QueryParser().parse(str(private_query)))`, we could see immediately the dissimilarities between the two. 
Besides, we could color the incorrect type nodes to make the changes more apparent. 

![wrong_parsed_query](https://user-images.githubusercontent.com/22152801/112838436-57121480-909d-11eb-9151-5736c53484be.png)

**I propose a function which permits to represent any parsed query or bits of query as a graphviz `Digraph`.**
Each node of the ast is then recursively visited to built the graph. Do you think that could be useful to integrate it in smartnoise ?

joshua-oss:
This is a great idea!

(Note that I have opened issue #346 to track the round-tripping issues, and we can discuss the visualization here).  Tagging @eedeleon into the discussion.

Where on the API should we put the functionality? One place to put this functionality could be as a method on `_ast.tokens.Sql`, from which all parsed AST nodes inherit.  So code like this would work:

```
query_parser = QueryParser(...)
q = query_parser.query('SELECT SUM(Foo) FROM Bar')
dot = q.visualize()

aggs = q.find_nodes(AggFunction)
for agg in aggs:
   dot = agg.visualize()
   ...
```

Any other options we should consider?

I assume that callers will want to call this API from console applications and Jupyter, with server and other applications being less common.  My understanding is that `Graphviz` in Jupyter will automatically convert to SVG to display in the browser, so callers will not need different code depending on the environment.  For callers who want to manually create PNG, etc, this seems easy enough that we wouldn't need helper methods.

vserret:
[...]

PR #349, opened 2021-04-06
add visualize method to class Sql

vserret:
The `visualize` method of the `Sql` class permits to construct the `Digraph` graphviz object of the Sql object, see [issue_345](https://github.com/opendp/smartnoise-sdk/issues/345).

Example: 
```
from opendp.smartnoise.sql.parse import QueryParser
from opendp.smartnoise._ast.ast import Query

query = "SELECT SUM(age) AS my_sum FROM pums.pums GROUP BY age"
parsed_query = QueryParser().query(query)
graph = parsed_query.visualize(color_types={Query:'red'}, n_trunc=30)
```

If your are working in a notebook, display the graph by doing: 
```
graph
```

Otherwise, use: 
```
graph.render('ast_digraph', view=True, cleanup=True)
```
In this case, the file ast_digraph.pdf if created. 

[ast_digraph.pdf](https://github.com/opendp/smartnoise-sdk/files/6264314/ast_digraph.pdf)

### DIFF
--- a/sdk/opendp/smartnoise/_ast/tokens.py
+++ b/sdk/opendp/smartnoise/_ast/tokens.py
@@ -129,6 +129,58 @@ class Sql:
         childnodes = [c.find_nodes(type_name, not_child_of) for c in sqlnodes]
         return nodes + flatten(childnodes)
 
+    def visualize(self, color_types={}, n_trunc=None):
+        """
+        Construct the AST graph of the object
+
+        Args:
+            color_type (optional): A dictionnary which contains the type and
+                the color of the nodes.
+                If no color is provided, the displayed node will be black.
+                Example: {str: 'red', Query: 'green'}
+            n_trunc: for visibility, truncate the expressions.
+                By default, there is no truncation.
+
+        Returns:
+            graphviz Digraph representation of 'parsed_query'.
+
+        """
+        def _label_node(expr, n_trunc):
+            str_expr = str(expr)
+            if n_trunc and len(str_expr) > n_trunc:
+                str_expr = str_expr[:n_trunc] + '...'
+            return f"{type(expr).__name__}: {str_expr}"
+
+        def _color_node(expr, color_types):
+            expr_type = type(expr)
+            if expr_type in color_types.keys():
+                return color_types[expr_type]
+            return 'black'
+
+        def _visit_nodes(node, path_to_node):
+            if not isinstance(node, Sql):
+                return
+            for child_node in node.children():
+                if child_node is not None:
+                    path_to_child_node = f"{path_to_node}.{str(child_node)}"
+                    graph.node(
+                        path_to_child_node,
+                        _label_node(child_node, n_trunc),
+                        color=_color_node(child_node, color_types),
+                    )
+                    graph.edge(path_to_node, path_to_child_node)
+                    _visit_nodes(child_node, path_to_child_node)
+
+        from graphviz import Digraph
+        graph = Digraph()
+        graph.node(
+            str(self),
+            _label_node(self, n_trunc),
+            color=_color_node(self, color_types),
+        )
+        _visit_nodes(self, str(self))
+        return graph
+
 
 class Seq(Sql):
     def __init__(self, seq):
--- a/sdk/opendp/smartnoise/_ast/tokens.py
+++ b/sdk/opendp/smartnoise/_ast/tokens.py
@@ -162,7 +162,7 @@ class Sql:
                 return
             for child_node in node.children():
                 if child_node is not None:
-                    path_to_child_node = f"{path_to_node}.{str(child_node)}"
+                    path_to_child_node = f"{path_to_node}.{type(child_node).__name__}_{str(child_node)}"
                     graph.node(
                         path_to_child_node,
                         _label_node(child_node, n_trunc),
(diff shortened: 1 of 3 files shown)


sdk/opendp/smartnoise/_ast/tokens.py
```diff
@@ -129,6 +129,58 @@ def find_nodes(self, type_name, not_child_of=None):
...
+
+        from graphviz import Digraph
+        graph = Digraph()
+        graph.node(
+            str(self),
+            _label_node(self, n_trunc),
+            color=_color_node(self, color_types),
+        )
+        _visit_nodes(self, str(self))
+        return graph
```
eedeleon, line 182: Is there any way to test this? I imagine the graph job is expected to have a certain shape. It would be worth having a minimal one node and maybe a 3 node option given it is recursive.

I am reading through more thoroughly now. The PR looks really simple with the current structure. Thanks for taking the time to make it read so well

sdk/opendp/smartnoise/_ast/tokens.py
```diff
@@ -129,6 +129,58 @@ def find_nodes(self, type_name, not_child_of=None):
...
+            str_expr = str(expr)
+            if n_trunc and len(str_expr) > n_trunc:
+                str_expr = str_expr[:n_trunc] + '...'
+            return f"{type(expr).__name__}: {str_expr}"
+
+        def _color_node(expr, color_types):
+            expr_type = type(expr)
+            if expr_type in color_types.keys():
+                return color_types[expr_type]
+            return 'black'
```
eedeleon, line 158: Can we make this a constant to explain why we pick this color as the default. I imagine it is the base color for the viz

sdk/opendp/smartnoise/_ast/tokens.py
```diff
@@ -129,6 +129,58 @@ def find_nodes(self, type_name, not_child_of=None):
...
+                        path_to_child_node,
+                        _label_node(child_node, n_trunc),
+                        color=_color_node(child_node, color_types),
+                    )
+                    graph.edge(path_to_node, path_to_child_node)
+                    _visit_nodes(child_node, path_to_child_node)
+
+        from graphviz import Digraph
+        graph = Digraph()
+        graph.node(
```
eedeleon, line 176: Can we add an example viz in the comment section here. Any future dev that contributes to this feature would benefit from an example viz

sdk/opendp/smartnoise/_ast/tokens.py
```diff
@@ -129,6 +129,58 @@ def find_nodes(self, type_name, not_child_of=None):
...
+                return color_types[expr_type]
+            return 'black'
+
+        def _visit_nodes(node, path_to_node):
+            if not isinstance(node, Sql):
+                return
+            for child_node in node.children():
+                if child_node is not None:
+                    path_to_child_node = f"{path_to_node}.{str(child_node)}_{str(id(child_node))}"
+                    graph.node(
```
eedeleon, line 166: rather than use the outer scope, can we pass graph as the first parameter? given the mutation it should work the same but be easier to follow

_visit_nodes(graph, node, path_to_node)

sdk/opendp/smartnoise/_ast/tokens.py
```diff
@@ -129,6 +129,58 @@ def find_nodes(self, type_name, not_child_of=None):
...
+
+        Returns:
+            graphviz Digraph
+
+        """
+        def _label_node(expr, n_trunc):
+            str_expr = str(expr)
+            if n_trunc and len(str_expr) > n_trunc:
+                str_expr = str_expr[:n_trunc] + '...'
+            return f"{type(expr).__name__}: {str_expr}"
```
joshua-oss, line 152: I think this syntax is only supported in Python 3.6+, but it should be OK, since our SQLite dependency for Pandas support now requires 3.6+.

tests/sdk/ast/test_visualize.py
```diff
@@ -0,0 +1,61 @@
...
+        private_reader = PrivateReader(reader, schema, 1.0)
+        inner, outer = private_reader.rewrite_ast(parsed_query)
+        graph = outer.visualize(n_trunc=30)
+        assert(isinstance(graph, Digraph))
+        #graph.render('ast_digraph', view=True, cleanup=True)
+        graph = inner.visualize(n_trunc=30)
+        assert(isinstance(graph, Digraph))
+        #graph.render('ast_digraph', view=True, cleanup=True)
+    def test_viz_child_nodes(self):
+        query = "SELECT AVG(age) AS my_sum FROM PUMS.PUMS GROUP BY age"
```
eedeleon, line 53: That was fast haha. I just mentioned tests would help. We need to update requirements.txt with graphviz @joshua-oss

joshua-oss (approved):
The PR looks great!  This will be very useful; thanks for the contribution!

sdk/opendp/smartnoise/_ast/tokens.py
joshua-oss: Sorry, just saw this.  I added 4 unit tests; currently just checking that the various calls return Digraphs, but the visualize lines can be uncommented to inspect the rendering, and everything matches as expected.  Note that the unit tests are causing integration tests to fail, due to graphviz not being present.  I just added graphviz to tests/requirements.txt to fix that

PR #351, opened 2021-04-08
modify the rewriter for using only AST

vserret:


sdk/opendp/smartnoise/sql/private_rewriter.py
```diff
@@ -94,44 +103,57 @@ def push_sum_or_count(self, exp, scope):
...
-            agg_exp.quantifier = None
-            agg_exp.expression = new_exp
+            new_name = scope.push_name(Column(exp.name))
+            exp.name = new_name
+
+        elif type(exp) is AggFunction:
+            exp = self.rewrite_agg_expression(exp, scope)
+
+        else:
+            for outer_col_exp in exp.find_nodes(Column, AggFunction):
```
eedeleon, line 144: This change is amazing. Thank you ! So much clearer + extra functionality

sdk/opendp/smartnoise/sql/private_rewriter.py
```diff
@@ -150,32 +172,32 @@ def query(self, query):
...
             child_scope.select(), query.source, query.where, query.agg, None, None, None
         )
         subquery = self.exact_aggregates(subquery)
-        subquery = [AliasedRelation(subquery, "exact_aggregates")]
-
+        subquery = [Relation(AliasedSubquery(subquery, Identifier("exact_aggregates")), None)]
         q = Query(select, From(subquery), None, query.agg, query.having, query.order, query.limit)
-
-        return QueryParser(self.metadata).query(str(q))
+        q.load_symbols(self.metadata)
```
eedeleon, line 184: Can load_symbols be part of the constructor? It seems off to me to have a mutation called after the constructor before returning. @joshua-oss  might have context here

sdk/opendp/smartnoise/sql/private_rewriter.py
```diff
@@ -257,10 +284,13 @@ def clamp_expression(self, ne, relations, scope, do_clamp=True):
...
-                        str(colname), minval, maxval
-                    )
-                    cexpr = Expression(clamped_string)
+                    when_min = WhenExpression(
+                        BooleanCompare(col, Op("<"), Literal(minval)), Literal(minval)
+                        )
+                    when_max = WhenExpression(
+                        BooleanCompare(col, Op(">"), Literal(maxval)), Literal(maxval)
+                        )
+                    cexpr = CaseExpression(None, [when_min, when_max], col)
```
joshua-oss, line 293: Nice!

sdk/opendp/smartnoise/sql/private_rewriter.py
```diff
@@ -207,9 +229,14 @@ def per_key_random(self, query):
...
                 NamedExpression(
-                    "row_num",
-                    Expression(
-                        "ROW_NUMBER() OVER (PARTITION BY {0} ORDER BY random())".format(key_col)
+                    Identifier("row_num"),
+                    RankingFunction(FuncName("ROW_NUMBER"),
+                                    OverClause(
+                                        Column(key_col),
+                                        Order([
+                                            SortItem(BareFunction(FuncName("RANDOM")), None)
```
joshua-oss, line 237: Great fix

sdk/opendp/smartnoise/sql/private_rewriter.py
joshua-oss: I will need to think about this for a bit.  The previous code was round-tripping through the parser, but the new code is more in line with what the original design intention was.  There is actually a note about this in the dev design document:

> The original intention was that AST parsing would branch based on presence of metadata, such that symbol resolution would be a separate and optional step. However, this code path is not well-factored or well-tested, and the current assumption is that metadata is always passed in to parser, and symbol resolution always happens automatically at parse.

So, the original intent was that ASTs can be constructed arbitrarily, and then symbol resolution can happen independently (e.g. clone an AST to multiple backend engines and do symbol resolution for each).  I switched to using `load_symbols` only internally, with external code required to pass through the parser, because I don't have a ton of confidence that the `load_symbols()` code path can handle everything an external party would throw at it.  As long as we are operating only on ASTs that we constructed ourselves, with unit tests, it's safe.  But I can imagine arbitrary externally-constructed ASTs that will blow it up in unexpected ways.

In this case, it may be OK, since we own the rewriter code, and we can be sure unit tests are passing.  The rewriter consumes ASTs that have already been through multiple stages of validation.

vserret: ok, I understand. 
So what dio you want me to do about it ?

vserret:
@eedeleon do you think this could be useful to add a pytest to check that the rewriter follows the grammar and that we have `rewrited_query == QueryParser(meta).query(str(rewrited_query))` ?

sdk/opendp/smartnoise/sql/private_rewriter.py
vserret: Could `metadata` be an optional argument of the `Query` constructor ? In that case, that would change nothing. 
Otherwise, if the AST blocks had type hints, then an external party would not be able to throw anything at it. What do you think ?

joshua-oss: @vserret, that sounds like a good idea.  Adding it to Query constructor would be an easy change that would let us make `load_symbols` more explicitly internal.  There is still the risk that people could modify the AST after construction, which would cause the symbols to be out of date.  But I think it's OK for that scenario to be out of scope for now.

eedeleon:
> @eedeleon do you think this could be useful to add a pytest to check that the rewriter follows the grammar and that we have `rewrited_query == QueryParser(meta).query(str(rewrited_query))` ?

Yes I do, this sort of end to end verification will catch regressions forsure

sdk/opendp/smartnoise/_ast/ast.py
```diff
@@ -22,7 +22,7 @@ def children(self):
 class Query(SqlRel):
     """A single query"""
 
-    def __init__(self, select, source, where, agg, having, order, limit) -> None:
+    def __init__(self, select, source, where, agg, having, order, limit, metadata=None) -> None:
```
joshua-oss, line 25: Nice!

joshua-oss (approved):
The PR looks really good!!

PR #355, opened 2021-04-20
fix grammar errors in ast blocks

vserret:
Some parts in the AST blocks does not follow the grammar rules

### DIFF
--- a/sdk/opendp/smartnoise/_ast/expression.py
+++ b/sdk/opendp/smartnoise/_ast/expression.py
@@ -5,6 +5,25 @@ from .expressions.logical import *
 from .expressions.sql import *
 
 
+EXPR_TYPE = Union[
+    "Expression",
+    "Column",
+    "ArithmeticExpression",
+    "CaseExpression",
+    "AllColumns",
+    "Literal",
+    "RankingFunction",
+    "BareFunction",
+    "RoundFunction",
+    "PowerFunction",
+    "AggFunction",
+    "MathFunction",
+    "IIFFunction",
+    "ChooseFunction",
+    "AliasedSubquery",
+    "NestedExpression",
+]
+
 class Expression(SqlExpr):
     """A bare expression with no name"""
 
@@ -58,7 +77,7 @@ class NamedExpression(SqlExpr):
     """An expression with optional name"""
 
     def __init__(
-        self, name: str, expression: Union["Expression", "Column", "ArithmeticExpression"]
+        self, name: Identifier, expression: EXPR_TYPE
     ) -> None:
         self.name = name
         self.expression = expression
@@ -70,9 +89,9 @@ class NamedExpression(SqlExpr):
             parts = self.expression.name.split(".")
             return parts[0] if len(parts) == 1 else parts[1]
         elif type(self.expression) is AllColumns:
-            return "???"
+            return Identifier("???")
         else:
-            return "???"
+            return Identifier("???")
 
     def type(self):
         return self.expression.type()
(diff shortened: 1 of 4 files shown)


sdk/opendp/smartnoise/_ast/expression.py
```diff
@@ -58,7 +77,7 @@ class NamedExpression(SqlExpr):
     """An expression with optional name"""
 
     def __init__(
-        self, name: str, expression: Union["Expression", "Column", "ArithmeticExpression"]
+        self, name: Identifier, expression: EXPR_TYPE
     ) -> None:
```
joshua-oss, line 81: This is so much better; thanks!

sdk/opendp/smartnoise/_ast/expressions/logical.py
```diff
@@ -85,7 +85,7 @@ def symbol(self, relations):
         return ColumnBoolean(self.expression.symbol(relations))
 
     def type(self):
-        return bool
+        return "boolean"
```
joshua-oss, line 88: Good catch

sdk/opendp/smartnoise/_ast/expression.py
```diff
@@ -70,9 +89,9 @@ def column_name(self):
             parts = self.expression.name.split(".")
             return parts[0] if len(parts) == 1 else parts[1]
         elif type(self.expression) is AllColumns:
-            return "???"
+            return Identifier("???")
```
joshua-oss, line 92: This line will fix the AST type of the `column_names` return value for `AllColumns` be consistent with the return type when called on any `NamedExpression` which has an alias name specified.  Both will return `Identifier` after this fix.  However, I believe we will still return `str` for unaliased columns (line 90 of this file), and for unaliased aggregates (using `symbol_name`).

I believe this is only used in the code as the key in the tuple of (name, symbol) under `m_symbols` hash table.  The code seems to be working OK currently, despite the hash table keys being a mix of `str` and `Identifier`, but this potentially makes us vulnerable to bugs.

This raises the question of what we want the return type of `column_name` to be.  The `column_name` defines the output column name which will be used when serializing the result rowset, and isn't part of the query AST.  When generating the output rowset, column names are always converted to `str` before use.  However, ensuring that all `column_name` outputs are `Identifier` could allow us to do some runtime validation to ensure that column names always adhere to the grammar definition of `IDENT`.

I don't have a strong preference either way.  Thoughts?

A second issue is that this particular line is underspecified, because `column_name` doesn't really have a meaning for `AllColumns`.  The output column names for any `SELECT *` will always expand to the full list of columns in the database, so it should be impossible to ever have a column name that is `???`.  During symbol resolution, `AllColumns` are expanded (metadata are required for the parser to know what column names comprise `*`), the symbol resolution will resolve the symbol as a single column of type `AllColumns`.  So we theoretically don't need this line.  It could serve as a failsafe if anyone tries to use a parsed query without symbols, but I'm wondering if we should just throw an exception in that case anyway?  I can't think of a case where calling `column_name` on `AllColumns` would not be an error.

vserret: You are right
I think this would be better to keep it like that for the moment and have latter one PR dedicated to avoiding  mixing `str` and `Identifier` because the changes would be trickier (for example to avoid to call `Identifier` on another `Identifier` in the Scope.push_name method).

joshua-oss: I just realized, we will never have the issue about `AllColumns` I raised above, because we don't support `SELECT *`.  And we do need to return `symbol_name` on `AllColumns`, because it's used for example in `SELECT COUNT(*)`, which resolves to `count_<AllColumns.symbol_name>`

PR #357, opened 2021-05-03
Working PR to update smartnoise-sdk docstrings to sphinx

lurosenb:
- MWEM and quail comments updated to be sphinx ready


lurosenb:
Just getting synthesizers ready to be added to the open-dp docs

eedeleon (approved):
This works for now. We should figure out if we want to use default docs or more human readable versions.

PR #358, opened 2021-05-14
Draft:Fix evaluate

vserret:
The `evaluate` method of the `CaseExpression` does not work. 
I do not understand the code which computes it in the `main` branch so maybe I'm missing something.

### DIFF
--- a/sdk/opendp/smartnoise/_ast/expressions/logical.py
+++ b/sdk/opendp/smartnoise/_ast/expressions/logical.py
@@ -236,12 +236,10 @@ class CaseExpression(SqlExpr):
         if self.expression is not None:
             # simple search
             for we in self.when_exprs:
-                match = BooleanCompare(self.expression, Op("="), we.expression).evaluate(bindings)
-                if match:
+                if BooleanCompare(self.expression, Op("="), we.expression).evaluate(bindings):
                     return we.then.evaluate(bindings)
         else:
             # regular search
-            print(bindings)
             for we in self.when_exprs:
                 if we.expression.evaluate(bindings):
                     return we.then.evaluate(bindings)
--- a/sdk/opendp/smartnoise/_ast/expressions/logical.py
+++ b/sdk/opendp/smartnoise/_ast/expressions/logical.py
@@ -233,19 +233,23 @@ class CaseExpression(SqlExpr):
         )
 
     def evaluate(self, bindings):
-        else_exp = self.else_expr.evaluate(bindings)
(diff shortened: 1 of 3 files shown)


joshua-oss:
I think this has always been buggy.  Let's add a unit test to verify it's working now.

Some of the expressions have not implemented `evaluate`, for example, `IIFFunction` and `ChooseFunction`.  Those can be implemented in a future PR.

One other note: `AggFunction.evaluate()` is currently just passing through to the inner `evaluate()`, which gives expected behavior if we assume that the `bindings` represent the results that have already been aggregated on the server.  However, if someone were expecting this to perform the actual aggregation, it would give unexpected results.  I assume this is OK, but I'm mentioning it in case you can think of problems with this behavior.

joshua-oss:
FWIW, I think [this unit test](https://github.com/opendp/smartnoise-sdk/blob/85dd0101af30fff593fdad2a8193868aaceedc37/tests/sdk/ast/test_logical.py) is the only place we are currently testing `evaluate()`.  The tests for `BooleanCompare` are excessive.  I think there are also a couple tests under `tests/query` that indirectly exercise the `evaluate()` code path for `ArithmeticExpression`

joshua-oss:
OK, I added some unit tests.  I noticed some inconsistency between handling string literals versus bound string values.  When I bind to a string, as in the test named `test_string_bound`, the call to `evaluate()` returns an unadorned string, as expected.  However, when I set the value to a string literal, as in `test_simple`, the call to `evaluate()` results in a string that includes the string quotes as part of the string.  This is not the correct behavior.

This isn't a bug in the `CaseExpression` code, BTW, but is something that likely affects all calls to `evaluate()` that can involve string literals.  The issue stems from the fact that class `Literal` always includes the quotes in its value, which is the right behavior when outputting to something that needs to adhere to the `Query` grammar.  And, since the method for `Literal.evaluate()` simply returns the value, the value has spurious quotes.  I can think of two options to fix it:

1. Change `Literal.evaluate()` to strip surrounding quotes.  This way, the `str()` method would work as expected when outputting grammar, and bindings would work.  Benefit of this approach is that it's simple to implement.
2. Update the `Literal` class to have a field to store the quote character, then change the parser to extract and store the quote character.  Modify the `str()` function on `Literal` to wrap the value in the detected quote character.  Benefit of this approach is that it might work better with varying quote characters across DB platforms.

Thoughts?

joshua-oss:
OK, I fixed the parser per proposal 2 above.  I'm approving, but please take a look and make sure my changes are OK

joshua-oss (approved):
Thanks for the fix!

Issue #359, opened 2021-05-14
Why clamping the grouping columns ? 


vserret:
The tau-thresholding makes sure that releasing groups is safe. So I think there is no use to clamp the grouping columns [here](https://github.com/opendp/smartnoise-sdk/blob/85dd0101af30fff593fdad2a8193868aaceedc37/sdk/opendp/smartnoise/sql/private_rewriter.py#L250). 

We could replace: 
```
select = [
                self.clamp_expression(ne, relations, child_scope, self.options.clamp_columns)
                for ne in query.select.namedExpressions
            ]
```
by: 
```
select = [
                self.clamp_expression(ne, relations, child_scope, 
               (self.options.clamp_columns is_grouping_expression(ne))
                for ne in query.select.namedExpressions
            ]
```
where we have to define `is_grouping_expression`. 
By doing that, we should be able to allow grouping by expressions with minimal changes.

joshua-oss:
That's an interesting observation.  My assumption was that grouping columns will typically be categorical, so won't be clamped.  But I could imagine someone (for example) doing `SELECT Salary, SUM(Bonus) GROUP BY Salary`, where `Salary` has a `minval` and `maxval`.  Is this what you were thinking of?  Or do you mean when grouping by some expression like `GROUP BY SomeVal % 5`?

In any case, I agree the grouping column needn't be clamped.

joshua-oss:
Issue fixed in #360

PR #360, opened 2021-05-17
Do not clamp grouping columns

vserret:
see issue #359

### DIFF
--- a/sdk/opendp/smartnoise/sql/private_rewriter.py
+++ b/sdk/opendp/smartnoise/sql/private_rewriter.py
@@ -32,7 +32,6 @@ from opendp.smartnoise._ast.ast import (
     Relation,
     SortItem,
     WhenExpression,
-    Sql,
 )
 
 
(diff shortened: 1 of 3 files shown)


sdk/opendp/smartnoise/sql/private_rewriter.py
```diff
@@ -247,45 +247,54 @@ def per_key_clamped(self, query):
...
             cols += [exp]
         for col in cols:
             colname = col.name
-            minval = None
-            maxval = None
-            sym = col.symbol(relations)
-            if do_clamp and sym.valtype in ["float", "int"] and not sym.unbounded:
-                minval = sym.minval
-                maxval = sym.maxval
-                if minval is None or sym.is_key:
```
vserret, line 274: I do not understand this line: why `maxval is None` does not appear here ?

sdk/opendp/smartnoise/sql/private_rewriter.py
```diff
@@ -247,45 +247,54 @@ def per_key_clamped(self, query):
...
+            if not do_clamp:
+                return None, None
+            if query.agg:
+                grouping_colnames = [col.name for col in query.agg.groupedColumns()]
+                if colname in grouping_colnames:
+                    return None, None
+            minval = None
+            maxval = None
+            sym = col.symbol(relations)
+            if sym.valtype in ["float", "int"] and not sym.unbounded:
```
vserret, line 272: I am not sure of this line
See my previous comment

sdk/opendp/smartnoise/sql/private_rewriter.py
joshua-oss: That's an oversight, it should be checking maxval.  I just pushed the change

sdk/opendp/smartnoise/sql/private_rewriter.py
joshua-oss: This change looks good.  We should add a unit test to make sure it's behaving as intended, and to protect against regressions in the future.

joshua-oss (approved):
Added unit test and approved, thanks for the fix!!

Issue #364, opened 2021-05-25
Implement evaluate on IIFFunction and ChooseFunction


joshua-oss:

joshua-oss:
Implemented in #365

PR #365, opened 2021-05-25
IIF and CHOOSE evaluate()

joshua-oss:
For #364

### DIFF
--- a/sdk/opendp/smartnoise/_ast/expressions/logical.py
+++ b/sdk/opendp/smartnoise/_ast/expressions/logical.py
@@ -287,8 +287,7 @@ class ChooseFunction(SqlExpr):
         # index in CHOICE is 1-based, not 0-based
         if len(self.choices) < idx or idx < 1:
             return Literal(None).evaluate(bindings) # NULL
-        else:
-            return self.choices[idx - 1].evaluate(bindings)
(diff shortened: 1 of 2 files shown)


sdk/opendp/smartnoise/_ast/expressions/logical.py
```diff
@@ -280,9 +280,17 @@ class ChooseFunction(SqlExpr):
...
-
     def children(self):
         return [Token("CHOOSE"), Token("("), self.expression, Token(","), self.choices, Token(")")]
+    def evaluate(self, bindings):
+        idx = int(self.expression.evaluate(bindings))
+        # index in CHOICE is 1-based, not 0-based
+        if len(self.choices) < idx or idx < 1:
+            return Literal(None).evaluate(bindings) # NULL
+        else:
+            return self.choices[idx - 1].evaluate(bindings)
```
vserret, line 291: The
```
if ...
   return ...
else:
   return ...
```
may be replaced by:
```
if ...
   return ...
return ...
```

vserret (approved):

sdk/opendp/smartnoise/_ast/expressions/logical.py
joshua-oss: Good call; fixed

Issue #344, opened 2021-03-23
Mixing COUNT and COUNT DISTINCT returns incorrect results


joshua-oss:
When a query requests both COUNT ALL and COUNT DISTINCT, e.g. `SELECT COUNT(pid), COUNT(DISTINCT pid)`, the two counts will be set to the same value.  This is because the private reader tries to be clever, and re-uses counts of the same key, but doesn't check for the quantifier.  The re-use code is intended to ensure that noise is not added to the same count multiple times, which would waste privacy budget and lead to potential inconsistency when (e.g.) the same count is used as a denominator for an AVG.  This code is activated only for counts of the `private_id`, since this is where the wasted budget is most likely to occur, due to a hidden count being added to all queries without `row_privacy`, which is necessary to support tau thresholding.  So this bug will manifest on any query that counts the `private_id` requesting both DISTINCT and ALL.  

This is not a privacy bug, but the error in accuracy could be large, and could change significantly depending on whether or not the query asks for more than one count with different quantifiers.

A simple patch is to comment out lines 229 and 230 in `private_reader.py`.  This disables the "cleverness", fixes the accuracy error, and doesn't have an impact on privacy budget, unless the key column is counted more than twice.

A better fix would be to disable this processing that is specific to counts on `private_id`, and implement a general-purpose re-use model that ensures that no statistics which ask for the same thing will add extra noise.  For example, if a query asks for `SUM (sales)`, and another asks for `AVG (sales) / 100` (which uses a sum), the private reader should automatically ensure that noise is only added to the sum once, and the noisy answer is used in computing both columns.

The proposed fix should use canonicalized representations of expressions in the AST to test for equivalence, and should support engine-specific canonicalization.  For example, `SUM(sales)` and `SUM(SALES)` will be equivalent expressions on some engines, but not others, depending on case-sensitivity and escaping rules.

joshua-oss:
This is fixed in #369

PR #369, opened 2021-05-31
Reuse matching expressions

joshua-oss:
Fix for #344 

Rewriter will properly re-use noise on expressions, and will keep `COUNT()` and `COUNT (DISTINCT)` separate.

Three unit tests:
1. Make sure keycount gets re-used for `COUNT (DISTINCT pid)` when reservoir sampling enabled
2. Make sure keycount gets-reused for `COUNT (*)` when `row_privacy` enabled
3. Make sure expressions are re-used when selecting `AVG`, `SUM`, and `COUNT` in the same `SELECT`

### DIFF
--- a/sdk/opendp/smartnoise/_ast/ast.py
+++ b/sdk/opendp/smartnoise/_ast/ast.py
@@ -254,7 +254,7 @@ class Relation(SqlRel):
             symbol
             for name, symbol in syms_b
             if (type(symbol) is TableColumn and symbol.compare.identifier_match(colname, name))
-            or name.lower() == colname.lower()
+            or (type(symbol) is not TableColumn and name.lower() == colname.lower())
         ]
         if len(syms_c) == 1:
             return syms_c[0]
--- a/sdk/opendp/smartnoise/_ast/ast.py
+++ b/sdk/opendp/smartnoise/_ast/ast.py
@@ -254,7 +254,7 @@ class Relation(SqlRel):
             symbol
             for name, symbol in syms_b
             if (type(symbol) is TableColumn and symbol.compare.identifier_match(colname, name))
-            or (type(symbol) is not TableColumn and name.lower() == colname.lower())
+            or name == colname
         ]
         if len(syms_c) == 1:
             return syms_c[0]
--- a/sdk/opendp/smartnoise/sql/private_rewriter.py
+++ b/sdk/opendp/smartnoise/sql/private_rewriter.py
@@ -160,6 +160,15 @@ class Rewriter:
         Validate().validateQuery(query, self.metadata)
 
         child_scope = Scope()
+
+        if self.options.row_privacy:
+            keycount_expr = AggFunction(FuncName("COUNT"), None, AllColumns())
+        else:
+            key_col = self.key_col(query)
+            keycount_expr = AggFunction(FuncName("COUNT"), Token("DISTINCT"), Column(key_col))
+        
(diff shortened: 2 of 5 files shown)


sdk/opendp/smartnoise/_ast/ast.py
```diff
@@ -254,7 +254,7 @@ def symbol(self, expression):
             symbol
             for name, symbol in syms_b
             if (type(symbol) is TableColumn and symbol.compare.identifier_match(colname, name))
-            or name.lower() == colname.lower()
+            or name == colname
```
joshua-oss, line 257: Line 256 enforces engine-specific comparison rules for `TableColumn`, and was working fine.  Line 257 is supposed to catch cases where the column being compared is not a `TableColumn`, but is instead a computed column like `ROW_NUMBER`.  However, this line was using `.lower()`, and would apply to `TableColumn` and non-`TableColumn` columns, which could result in a `TableColumn` correctly failing to get a match in the test above, and then erroneously matching on the next line.  Switching to use case-sensitive is safer.

sdk/opendp/smartnoise/sql/private_rewriter.py
```diff
@@ -182,19 +191,10 @@ def query(self, query):
...
-            key_col = self.key_col(query)
-            keycount_expr = AggFunction(FuncName("COUNT"), Token("DISTINCT"), Column(key_col))
-            child_scope.push_name(keycount_expr.expression)
-
         for ne in query.select.namedExpressions:
             child_scope.push_name(ne.expression)
 
-        keycount = NamedExpression(Identifier("keycount"), keycount_expr)
-
-        select = [keycount] + [ne for ne in query.select.namedExpressions]
```
joshua-oss, line 197: Prepending `[keycount]` meant that the keycount was not registered in the child scope of the outer call (`query()`), which prevented the rewriter from reusing keycount expressions.

sdk/opendp/smartnoise/sql/private_rewriter.py
```diff
@@ -160,6 +160,15 @@ def query(self, query):
...
         child_scope = Scope()
+
+        if self.options.row_privacy:
+            keycount_expr = AggFunction(FuncName("COUNT"), None, AllColumns())
+        else:
+            key_col = self.key_col(query)
+            keycount_expr = AggFunction(FuncName("COUNT"), Token("DISTINCT"), Column(key_col))
+        
+        child_scope.push_name(keycount_expr, "keycount")
+
```
joshua-oss, line 171: We move the keycount to the `query()` function and push to child scope

tests/sdk/query/test_validate_pums.py
```diff
@@ -38,6 +39,7 @@
 bad_files = [f for f in validate_files if "_fail" in f]
 
 metadata = CollectionMetadata.from_file(meta_path)
+metadata.compare = PostgresNameCompare()
```
joshua-oss, line 43: This was necessary after fixing the issue in line 257 of ast.py.  We have a unit test to make sure we get a validation error on `SELECT SUM(Age + Income) FROM PUMS.PUMS`.  This was passing before the bug fix, but failed after the bug fix, because it was using `PandasReader`, which is case sensitive.  So, this unit test should have been failing.  Attaching the `PostgresNameCompare` switches the test to be case-insensitive, so the tests succeed.

vserret (approved):

Issue #325, opened 2020-12-15
Allow Passing of Established SQL Connections


joshua-oss:
Our current design tightly couples information that we need for differential privacy (such as NameCompare) with engine-proprietary connection information that is used to establish a database connection.  This limits callers to connection types that we support (currently only username/password to an IP host and port).  This limitation is unnatural and unnecessary, since callers are accustomed to managing connections outside the scope of calling APIs.

The proposal is to update the API to allow callers to pass in established connections.  This has several advantages:
* Works seamlessly with external middleware such as gateways and load balancers
* Works seamlessly with more robust authentication techniques like certificate-based authentication
* Minimizes requirement for caller to learn new connection API
* Allows caller to externally manage transactions, including nested and distributed
* Facilitates transparent attach of engine-specific behavior we need

# Code Samples

The following code samples use a static factory method.  This is not the only possible design.

## Per-Engine Factory

```
import pyscopg2
conn = psycopg2.connect(dbname='PUMS', user='pums', sslmode='verify-full')

reader = PostgresReader.from_connection(conn)

priv_reader = PrivateReader(reader, meta)
```

This example improves on the current design, because the caller can use proprietary connection properties that apply only to the Postgres implementation of DP-API.  The `PostgresReader` uses the supplied connection, and simply serves as an access point for the `PostgresNameCompare` and `PostgresSerializer`.

Implementation cost is low.

## Generic Factory

We already have dynamic switching of `NameCompare` functionality based on the `engine` property in metadata.  This was a requirement, because we need to be able to parse and load symbols without an active connection.

```
import pyscopg2
conn = psycopg2.connect(dbname='PUMS', user='pums', sslmode='verify-full')

reader = SqlReader.from_connection(conn, "postgres")

priv_reader = PrivateReader(reader, meta)
```

The advantage of this approach is that the caller doesn't need to import classes or memorize the class names.  All connections are wrapped directly from SqlReader object.  

A disadvantage is that callers may use mismatching engine names (e.g. pass a SQL Server connection to the facory, while asking for "postgres" overrides).  This could be desirable, for example, if a caller has an unsupported 3rd-party database connection that is known to be compatible with an engine we support.

Implementation cost for this would be low.

## Generic Factory, Simplified

```
import pyscopg2
conn = psycopg2.connect(dbname='PUMS', user='pums', sslmode='verify-full')

reader = SqlReader.from_connection(conn)

priv_reader = PrivateReader(reader, meta)
```

[...]

PR #375, opened 2021-06-14
Sql conn passthru

joshua-oss:
Readers can now be created from an established connection.

```
conn = pyodbc.connect(dsn)
reader = SqlReader.from_connection(conn, "Postgres")
```

This addresses issue #325.

Added unit tests to allow testing against locally installed databases with credentials stored in local keyring.


lurosenb (commented):
This is awesome Joshua - however, it looks like its failing to pass the azure-pipeline tests (even if it is passing the github actions passthrough tests) as they are not installed. (It's giving: "ModuleNotFoundError: No module named 'pyodbc', 'psycopg2'")

Do these need to be added to the tests/requirements.txt? Alternatively, I can disable the azure-pipelines build for now, as we are transitioning over to Github actions in the longterm - let me know : )

joshua-oss:
> This is awesome Joshua - however, it looks like its failing to pass the azure-pipeline tests (even if it is passing the github actions passthrough tests) as they are not installed. (It's giving: "ModuleNotFoundError: No module named 'pyodbc', 'psycopg2'")
> 
> Do these need to be added to the tests/requirements.txt? Alternatively, I can disable the azure-pipelines build for now, as we are transitioning over to Github actions in the longterm - let me know : )

I recommend that we disable azure-pipelines and start using GH Actions exclusively.  I don't have access to fix the azure-pipelines builds, but I was able to port all of the SDK tests GH Actions and updated to make sure they install the right dependencies.  Note that azure-pipelines was failing even before this PR.

GH Actions also tests documentation build on every pull request, and there's a manually launched action to test notebooks.

The service tests haven't been ported over yet, so we should do that soon, but we might want to do that when we move service to be a sibling of SDK.

lurosenb:
> > This is awesome Joshua - however, it looks like its failing to pass the azure-pipeline tests (even if it is passing the github actions passthrough tests) as they are not installed. (It's giving: "ModuleNotFoundError: No module named 'pyodbc', 'psycopg2'")
> > Do these need to be added to the tests/requirements.txt? Alternatively, I can disable the azure-pipelines build for now, as we are transitioning over to Github actions in the longterm - let me know : )
> 
> I recommend that we disable azure-pipelines and start using GH Actions exclusively. I don't have access to fix the azure-pipelines builds, but I was able to port all of the SDK tests GH Actions and updated to make sure they install the right dependencies. Note that azure-pipelines was failing even before this PR.
> 
> GH Actions also tests documentation build on every pull request, and there's a manually launched action to test notebooks.
> 
> The service tests haven't been ported over yet, so we should do that soon, but we might want to do that when we move service to be a sibling of SDK.

Gotcha! Makes total sense - I believe I have disabled the azure-pipelines build.

lurosenb (approved):

PR #378, opened 2021-06-18
draft:query.load_symbols to set query.row_privacy for any query

vserret:
If the query has no table column (eg: `query = SELECT COUNT(*) FROM census.census` or `SELECT COUNT(1) FROM census.census`), if you do:
```
parsed_query = QueryParser().query(query)
parsed_query.load_symbols(metadata)
```
Then `parsed_query = False` even if `metadata`specifies the contrary. 

In this MR, I propose to set `query.row_privacy` even if there is no TableColumn in the query.

### DIFF
--- a/sdk/opendp/smartnoise/_ast/ast.py
+++ b/sdk/opendp/smartnoise/_ast/ast.py
@@ -67,25 +67,6 @@ class Query(SqlRel):
         self.m_symbols = symbols
         self.m_sym_dict = {}
 
-        """
-        tables = []
-        for name, symbol in self.m_symbols:
-            for table in symbol.find_nodes(TableColumn):
-                tables.append(table)
-            if name == "???":
-                continue
-            if name in self.m_sym_dict:
-                raise ValueError("SELECT has duplicate column names: " + name)
-            self.m_sym_dict[name] = symbol
-        print(tables)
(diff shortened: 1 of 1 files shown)


sdk/opendp/smartnoise/_ast/ast.py
```diff
@@ -67,16 +67,17 @@ def load_symbols(self, metadata):
...
-                tables.append(table)
             if name == "???":
                 continue
             if name in self.m_sym_dict:
                 raise ValueError("SELECT has duplicate column names: " + name)
             self.m_sym_dict[name] = symbol
 
+        tables = []
+        for t in self.find_nodes(Table):
+            tables.append(t.m_symbols[0][1])
```
vserret, line 79: I do not understand why `row_privacy` is loaded in `TableColumn` and not in `Table` since in `CollectionMetadata`, `row_privacy` is set for the table and not the line. 

Maybe I am missing something, but this change should fix https://github.com/opendp/smartnoise-sdk/issues/380 in the case were the tables are not empty. Should I raise an error otherwise ?

joshua-oss (approved):
This implementation is much better.  I don't even recall why I was using `TableColumn` here.  Thanks for fixing this.

PR #379, opened 2021-06-18
Removing azure pipeline yaml etc.

lurosenb:
Let's remove the devops and officially transition to github actions!


joshua-oss (approved):
This is great, thanks for the fix!

PR #381, opened 2021-06-18
Removed service from repo and tests, datasets are up one level

lurosenb:


joshua-oss (approved):
Looks good; thanks for doing this.

PR #393, opened 2021-08-30
Added attack notebook demonstrating database reconstruction attack

lurosenb:
Using IBM's ART library, demonstrate the efficacy of differentially private synthetic data at creating a dataset with high utility that preserves user privacy.

### DIFF
--- a/dpsdgym/example_attacks/database_reconstruction_attack.ipynb
+++ b/dpsdgym/example_attacks/database_reconstruction_attack.ipynb
@@ -1,0 +1,371 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "aa4b55d6",
+   "metadata": {},
+   "source": [
+    "# Preventing a Database Reconstruction Attack with Differentially Private Synthetic Data\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


joshua-oss (approved):
I was able to get all unit tests and CI tests passing on Ubuntu and Mac, though only on Python 3.7 and 3.8 (conda).  It appears that the default pip in 3.9 has recently changed for the worse, so it could take awhile to debug.  The notebook doesn't seem to work, because installing imblearn uninstalls sklearn, and ctgan and imblearn require different versions of sklearn.  On Windows, ctgan's requirement for an old version of pandas is conflicting with other dependencies, and I have been unable to get the dependencies installed in any version of python.

I recommend approving and merging this, since it gets the install, unit, and CI working on Linux and Mac again.  It feels like this is only working through pure luck, though, since the new pip resolver is particularly awful, and I expect more breaking changes.  I'll focus on reducing or compartmentalizing dependencies.  Porting the functionality of ctgan to run out of our repo with support for the newest numpy, sklearn, and pandas would undoubtedly help, but is unlikely to be the whole solution.

PR #394, opened 2021-09-06
Extend grammar

vserret:
- Extend the definition of an identifier, so it can handle extra characters (for example, French or Japanese words): 
```
SELECT "âge" FROM BAZ;
SELECT "検知日時" FROM BAZ;
SELECT 検知日時 FROM BAZ;
```

- Extend the numeric functions. Add `LN`, `CEIL`, `ATANH`, `LOG2`
- Add `STDEV`(https://docs.microsoft.com/en-us/sql/t-sql/functions/stdev-transact-sql?view=sql-server-ver15)


vserret:
Other functions could be added in the grammar (`TRUNC` / `TRUNCATE`, `ATAN2`, `LOG(base, number)`, `POWER(base, exponent)`, `CBRT`, `RADIANS`) but these additions require to add new AST. Should I add them in this MR or in another one ?

joshua-oss (approved):
This is awesome, thanks for adding these!  I merged back some CI fixes from master to ensure the unit tests are all passing.

The other additions you mentioned also sound good, let's merge this PR to master and make those in another PR.

PR #417, opened 2021-10-30
Small synthesizer fixes

lurosenb:
Made changes to fix odd mwem sampling, allow numpy as data type for pytorch_dp_synths:

- [x] MWEM trained from a dataframe seems to require sample() to use the same number of rows as training, so it needs numpy array
- [x] * MWEM seemingly requires split_factor to be set; it throws a strange error when unset
- [x] * PATE-CTGAN and DP-CTGAN don’t seem to allow numpy arrays, but work fine with pandas


joshua-oss (approved):
Looks great; thanks!

PR #425, opened 2021-11-13
Check input types to categorical columns

joshua-oss:
Alert caller if continuous values are passed to categorical column


lurosenb (approved):
Awesome - nice fixes/docs!

PR #427, opened 2021-11-15
dpctgan performance fix

AprilXiaoyanLiu:
fix 1: update default pac number to 1 
fix 2: update monkeypatch to coordinate with opacus 0.14.0, now wasserstein loss should work with opacus 0.14.0


lurosenb (approved):
So simple - just replicated the results on my machine - I'm good to check this in, @joshua-oss ?

joshua-oss (approved):
Nice!

Issue #432, opened 2021-11-19
DPCTGAN and PATECTGAN should reject continuous columns


joshua-oss:
For the CTGAN-based synthesizers, using GMM in preprocessing is not safe.  We document that continuous columns are not supported, and we throw an exception when callers pass continuous-valued columns to `PytorchDPSynthesizer` with `GeneralTransformer`.  However, there are 2 cases where the checks can fail, leading the synthesizers to silently use a GMM on continuous columns.

1. If the caller bypasses the `PytorchDPSynthesizer` wrapper and calls `DPCTGAN` or `PATECTGAN` directly, `GeneralTransformer` is not used, and SDV's `DataTransformer` is used instead.  This isn't an intended calling pattern, but people will do this, and it should be safe.
2. If the caller sets transformer to `None` on `PytorchDPSynthesizer`, the checks that are performed in `GeneralTransformer` will be skipped.  This generally leads to an exception, but could lead to one of the base synthesizers using SDV's GMM.

The simplest fix would be to modify the step right after `DataTransformer` gets fit:

```python
        self._transformer = DataTransformer()
        self._transformer.fit(data, discrete_columns=categorical_columns)
        for tinfo in self._transformer._column_transform_info_list:
            if tinfo.column_type == "continuous":
                raise ValueError("We don't support continuous values on this synthesizer.  Please discretize values.")
```

This would result in both synthesizers using only the one-hot-encoder from DataTransformer (which is intended).  We should probably just eliminate this dependency and use one-hot-encoder in GeneralTransformer, where we can ensure that all reversible transforms are privacy-safe.  However, the code that creates conditional vectors depends on the metadata format returned by DataTransformer's `fit`, so ripping it out will not be trivial.

joshua-oss:
Fixed

PR #434, opened 2021-11-19
CTGANs should reject continuous columns

joshua-oss:
Addresses #432


joshua-oss:
This will be a breaking change for anyone who was passing in columns that weren't listed in `categorical_columns`, which could be pretty common, especially with QUAIL

lurosenb:
Hmmm yeah, important fix though. Let's check this in, and discuss adding some built in continuous bucketing features in a followup PR that are disabled by default, to help users preprocess to categorical...thoughts?

joshua-oss:
> Hmmm yeah, important fix though. Let's check this in, and discuss adding some built in continuous bucketing features in a followup PR that are disabled by default, to help users preprocess to categorical...thoughts?

Agreed.  We should push out the bug fix ASAP, but then loop back and address continuous columns.  DPGAN and PATEGAN both handle continuous columns fine, because they don't use `DataTransformer`.  Simply disabling `DataTransformer` in our `CTGAN` synthesizers would allow us to pass through continuous columns again.  However, that could take awhile, because the construction of conditional vector is tightly coupled with `DataTransformer's` metadata.  And the CTGAN paper makes it seem like this "mode-specific normalization" (4.2) is important to the ability to concatenate continuous and categorical columns.

If we rip out `DataTransformer` and allow continuous columns back in, we'll be punting the preprocessing to the analyst.  We can give some code samples of how to safely preprocess, and eventually add safe continuous preprocessing to the API.  Min-max normalization is easy when bounds are known for the column.  This is a common requirement for computing sensitivity of DP queries, but would need to be exposed on the API somehow.  Also, section 4.2 of the CTGAN paper specifically mentions min-max normalization as undesirable.  So we will need to do some testing.

Besides min-max, an ordinal encoder probably makes sense.  SDV has one that uses category option frequency to choose numerical representation of the ordinal, and that would not be differentially private.  But something simple like using sequential integers and then min-max normalizing would work.

AprilXiaoyanLiu:
> > Hmmm yeah, important fix though. Let's check this in, and discuss adding some built in continuous bucketing features in a followup PR that are disabled by default, to help users preprocess to categorical...thoughts?
> 
> Agreed. We should push out the bug fix ASAP, but then loop back and address continuous columns. DPGAN and PATEGAN both handle continuous columns fine, because they don't use `DataTransformer`. Simply disabling `DataTransformer` in our `CTGAN` synthesizers would allow us to pass through continuous columns again. However, that could take awhile, because the construction of conditional vector is tightly coupled with `DataTransformer's` metadata. And the CTGAN paper makes it seem like this "mode-specific normalization" (4.2) is important to the ability to concatenate continuous and categorical columns.
> 
> If we rip out `DataTransformer` and allow continuous columns back in, we'll be punting the preprocessing to the analyst. We can give some code samples of how to safely preprocess, and eventually add safe continuous preprocessing to the API. Min-max normalization is easy when bounds are known for the column. This is a common requirement for computing sensitivity of DP queries, but would need to be exposed on the API somehow. Also, section 4.2 of the CTGAN paper specifically mentions min-max normalization as undesirable. So we will need to do some testing.
> 
> Besides min-max, an ordinal encoder probably makes sense. SDV has one that uses category option frequency to choose numerical representation of the ordinal, and that would not be differentially private. But something simple like using sequential integers and then min-max normalizing would work.

I agree that we can separate transformer out and make it DP, this can be both applied to DPGAN, PATEGAN, DPCTGAN AND PATECTGAN, and further synthesizers. DataTransformer has some meta info that conditional generator use, but should be easy to make it happen, it just try to store which column is which so can generate cond vector later.

Rajneesh-Tiwari:
Following up on this to see if there's been any resolution?

joshua-oss:
> Following up on this to see if there's been any resolution? 

The CTGANs currently reject continuous columns, requiring the caller to discretize/bin continuous columns before training.

We are planning to bring in support for continuous columns some time in January, with the following design:

1.  We will automatically apply a DP standard scaler to continuous columns.
2. Caller can specify that a continuous column has already been normalized in a way that doesn't impact privacy, in which case we will accept it and do no preprocessing
3. (tentative) If the caller passes in metadata from the SQL layer, specifying public upper and lower bounds, we will automatically apply a non-DP standard scaler

The Bayesian GMM will not be supported, since this could leak privacy.  If the continuous column is strongly multimodal, discretizing the values will probably work best.

Feedback is very welcome.

PR #435, opened 2021-11-20
New column metadata properties

joshua-oss:
* `sensitivity` can be used to override bounds-based sensitivity calculation
* `nullable=False` specifies that a column does not have missing values
* `missing_value` allows a fixed value to replace missing values via SQL `COALESCE`
* removed superfluous `neighboring` flag on `Privacy` object


sql/Metadata.md
```diff
@@ -55,10 +55,13 @@ These overrides should be used with caution, because they may affect privacy if
...
+* `type`: Required. The type attribute indicates the simple type for all values in the column. Type may be one of “int”, “float”, “string”, “boolean”, or “date”. The “date” type includes date or time types.  If type is set to "unknown", the column will be ignored by the system.
 * `private_key`: Boolean.  Default is `False`.  indicates that this column is the private identifier (e.g. “UserID”, “Household”).  This column is optional.  Only columns which have private_id set to ‘true’ are treated as individuals subject to privacy protection.
 * `lower`: Valid on numeric columns.  Specifies the lower bound for values in this column.
 * `upper`: Valid on numeric columns.  Specifies the upper bound for values in this column.
+* `nullable`: Boolean.  Default is `True`.  Indicates that this column can contain null values.  If set to `False`, the system will assume that all values are set.  This is useful when the data curator knows that all values are set, and will allow some budget to be preserved by sharing counts across columns.
```
Shoeboxam, line 62: Another approach is to add an impute_constant metadata field, and to adjust the query to use IFNULL.

sql/Metadata.md
```diff
@@ -55,10 +55,13 @@ These overrides should be used with caution, because they may affect privacy if
...
+* `missing_value`: A value of the same type as the `type` for this column.  Default is `None`.  If set, the system will replace NULL with the specified value, ensuring that all values are set.  If set, `nullable` will be treated as `False`, regardless of its value.
+* `sensitivity`: The sensitivity to be used when releasing sums from this column.  Default is `None`.  If not set, the system will compute the sensitivity from upper and lower bounds.  If `sensitivity` is set, the upper and lower bounds will be ignored for sensitivity, and this value will be used.  The upper and lower bounds will still be used to clamp the columns. If this value is set, and no bounds are provided, the metadata must specify `clamp_columns` as `False`. Note that counts will always use a sensitivity of 1, regardless of the value of this attribute.
```
Shoeboxam, line 64: Perhaps `sum_sensitivity` is better? Up to you.

vserret (approved):

sql/Metadata.md
joshua-oss: We've indiscriminately called it sensitivity everywhere else, meaning sensitivity used by sums.  Are there any scenarios where `count_sensitivity` or `mean_sensitivity` would be useful?  I guess I could imagine someone setting either one in cases with a fixed `n`, but unsure if it's a real scenario.

Shoeboxam: Sure, that's fine. What's the point of distinguishing what type of sensitivity you have, when there's only one kind of sensitivity! I understand.

joshua-oss: Yes, I'm concerned that someone might think there is a `count_sensitivity` and `mean_sensitivity` if `sum_sensitivity` is exposed.  If there's a chance someone would want `count_sensitivity` and/or `mean_sensitivity`, I would rather use `sum_sensitivity` and implement the other two now, since it would be a breaking change to rename `sensitivity` later.  It wouldn't be hard to implement the other two; I'm just not sure what people will want, or if it would be confusing.

For example, I could imagine someone having a fixed `n` deciding to use the sensitivity for the mean that's obtained by dividing sum sensitivity by n.  I don't know if this would ever be a good idea, but it seems like there could be cases where it makes sense.  I don't know enough to have a strong opinion.

Since we operate on aggregates, that's equivalent to setting `sum_sensitivity` and passing in the result of the SQL `AVG` to `pre_aggregated`, while converting the outer query to use a `SUM` instead of `AVG`.  If the caller passed in `mean_sensitivity` instead, the rewriter could treat it like a `sum_sensitivity` under the covers, but we'd have to make a decision about what to require in `pre_aggregated`. For example, we could:

A) Require count and sum in `pre_aggregated` (so it would be consistent with the most common cases when `mean_sensitivity` is not set).  This would be easy enough to implement: just compute the exact average from count and sum, and use the sensitivity specified by the user to add noise to the exact mean.

B) The rewriter could recognize that mean sensitivity is available, and could pull back exact mean (rather than a count and sum) as a single column.  This might match some users' expectations.  In fact, some callers might have only an exact mean, based on the system generating the data.  Obviously, if the caller knows an exact mean and an exact `n`, they implicitly know the sum.  But maybe it would be confusing to require conversion to sum?

The situation for `count_sensitivity` is easier, but we would still want to think about what happens if (e.g.) someone does an average over a column with `count_sensitivity` specified.

So it's basically, "one kind of sensitivity (sum) and make everyone use that as a workaround", risking a future breaking change; or, "figure out some sensible heuristics for all of the different sensitivities".  Attempting the latter imperfectly isn't terrible, since callers can always ignore our heuristics and use sum for everything.  I think it depends on how important we expect non-sum overrides to be.

PR #440, opened 2022-01-03
Add DP Standard Scaler to preprocess continuous columns in DPCTGAN

AprilXiaoyanLiu:
- add DP Standard Scaler to  preprocess continuous columns in DPCTGAN (preprocessors/data_transformer), though this will not capture multi-mode scenario but has reasonable performance, so we can put this as a candidate to preprocess continuous variable while in the meantime to work on DP GMM. 
- update dpctgan to use DP Standard Scaler to preprocess continuous columns
- to-do: update patectgan, pategan, ctgan 
- requires diffprivlib >= 0.4.1


synth/snsynth/preprocessors/data_transformer.py
```diff
@@ -0,0 +1,196 @@
...
+
+            self.output_info_list.append(column_transform_info.output_info)
+            self.output_dimensions += column_transform_info.output_dimensions
+            self._column_transform_info_list.append(column_transform_info)
+
+    def _transform_continuous(self, column_transform_info, raw_column_data):
+
+        scaler = column_transform_info.transform
+        normalized_values = scaler.transform(raw_column_data)
+        normalized_values = normalized_values/4
```
lurosenb, line 99: Why are we dividing by 4 here and then undoing that later?

AprilXiaoyanLiu: original CTGAN paper was done similarly but with GMM. I also experimented with it, if we divide by 4 and then undoing it later, the distribution of fake data way more aligns with the original data distirbution

lurosenb: Huh - that's odd!

lurosenb (approved):
Really nice addition! Should add to the other synthesizers as well

Issue #439, opened 2021-12-27
Potentially infinite loop and privacy issue in MWEM implementation?


shlomihod:
In the MWEM implementation, there is a while loop for making sure that no query is not chosen more than once (in the exponential mechanism part)

https://github.com/opendp/smartnoise-sdk/blob/3653e510d70f483e339219efa56e68e6ec5d5642/synth/snsynth/mwem.py#L245-L252

There are two issues with that:
1. It might lead to an infinite loop if the worst-performing query is chosen repeatedly, e.g., if the error is very high compared to the rest (this is how I actually discovered that after I run it on my data with eps=4).
2. I'm not sure if it is still DP with the same epsilon; or at least the original paper proof is without this step.

I checked two additional implementations for reference, and this step does not appear there:

1. Julia implementation by Moritz Hardt (co-author of the paper that presented the MWEM algorithm): https://github.com/mrtzh/PrivateMultiplicativeWeights.jl/blob/master/src/mwem.jl
2. Python implementation by Liu, Vietri and Wu (https://arxiv.org/pdf/2106.07153.pdf):  https://github.com/terranceliu/iterative-dp/blob/e5d7d28539fae0d2fed794168a78f97c869c0dbe/mwem.py#L39

I'd love to create a pull request if a change needs to be made.

joshua-oss:
Thanks for flagging this issue.  Adding @lurosenb to discuss.

It does indeed look like an infinite loop is possible.  Line 249 is ensuring that we do not measure the same query more than once (i.e. it loops until it gets a query that has not already been measured).  This line of code is analogous to the reference implementation in Figure 8, the last page of the MWEM paper [1] starting with the comment `// determine a new query to measure, rejecting prior queries`.  The paper's reference implementation will also have an infinite loop if the number of possible queries is less than the number of iterations.  Line 234 of our code should be raising a warning when this is the case.

The infinite loop should only possible when the number of possible queries is less than the number of iterations.  Until all queries have been measured, there will be a non-zero (though perhaps small) chance that a remaining query will be selected.  So the selection loop will eventually terminate.  However, if all queries have been measured, the loop will never terminate.

Rather than using an unbounded loop to reject prior queries, the Julia implementation sets the differences of prior queries to 0.0, to make sure they don't get chosen in the argmax:
https://github.com/mrtzh/PrivateMultiplicativeWeights.jl/blob/master/src/mwem.jl#L34

[...]

PR #442, opened 2022-01-05
Fix issue with MWEM infinite loop

lurosenb:
Simple temporary patch to ensure that MWEM terminates with an error when it cannot measure as many queries as it has iterations

Thanks @shlomihod for pointing this out in Issue #439 !


joshua-oss (approved):
Looks good; thanks!!

PR #454, opened 2022-03-13
Added the MST wrapper to snsynth. 

lurosenb:
Wrapper for code from https://github.com/ryan112358/private-pgm/blob/master/mechanisms/mst.py .

Tao et. al report on improvements of MST over current Smartnoise synthesizers in "Benchmarking Differentially Private Synthetic Data Generation Algorithms" https://arxiv.org/abs/2112.09238. In advance of synthetic data factory, we should support this method, for comparison and also because it is quite good in practice.

Moved over proper license and support files. Added smoke test.


joshua-oss (approved):
Looks good; this is really important work!

Issue #457, opened 2022-03-22
Column names with hyphens are unsupported


JMMarchant:
It appears that using `snsql` when the column names have hyphens in them is unsupported and there doesn't seem to be any way to escape/quote the column name as you would in a normal SQL query to ensure it doesn't get incorrectly split. Consider the following MWE:

```python
import pandas as pd
import snsql

# Dataframe with hyphenated and non-hyphenated column names
df = pd.DataFrame.from_dict(
    data={
        "age": {0: 25, 1: 38, 2: 28, 3: 44, 4: 18},
        "capital-gain": {0: 0, 1: 0, 2: 0, 3: 7688, 4: 0},
        "occupation": {
            0: "Machine-op-inspct",
            1: "Farming-fishing",
            2: "Protective-serv",
            3: "Machine-op-inspct",
            4: "?",
        },
    }
)

# Queries attempting to use hyphenated column
query_unquoted = (
    "SELECT occupation, AVG(capital-gain) FROM schema.table GROUP BY occupation"
)
query_quoted = (
    'SELECT occupation, AVG("capital-gain") FROM schema.table GROUP BY occupation'
)
query_backticks = (
    "SELECT occupation, AVG(`capital-gain`) FROM schema.table GROUP BY occupation"
)

# Query using non-hyphenated column
query_no_hyphen = "SELECT occupation, AVG(age) FROM schema.table GROUP BY occupation"

metadata = {
    "": {
        "schema": {
            "table": {
                "age": {
                    "type": "int",
                    "lower": 0,
                    "upper": 100,
                },
                "capital-gain": {
                    "lower": 0,
                    "upper": 99999,
                    "type": "int",
                },
                "occupation": {
                    "type": "string",
                },
                "row_privacy": True,
                "rows": 5,
            },
        },
    },
}

# Create reader from DataFrame
privacy = snsql.Privacy(
    epsilon=1.0,
    delta=1e-5,
)
reader = snsql.from_df(df, privacy=privacy, metadata=metadata)

# Each of these fails in a different way
# ValueError: Symbol could not be found in any relations: capital
reader.execute(query_unquoted)
# ValueError: Symbol could not be found in any relations: "capital-gain"
reader.execute(query_quoted)
# ValueError: Lexer error at line 1 column 23.  Message: token recognition error at: '`capital-'
reader.execute(query_backticks)

# Runs fine
reader.execute(query_no_hyphen)
```

[...]

Issue #456, opened 2022-03-13
Cannot find metadata when using `FROM database.schema.table`


oskarkocol:
`FROM database.schema.table` will fail to find metadata for a dataset because `metadata.py` only matches statements with `schema.table`:
https://github.com/opendp/smartnoise-sdk/blob/186b81721c0f5c9ef9028e56a5ac142e9b362602/sql/snsql/metadata.py#L26-L29
**Possible solution**: to add an additional `if` statement that matches three elements from an identifier like `database.schema.table`

joshua-oss:
I think the issue here is that the metadata specification assumes one database per metadata file.  The metadata file can have multiple top-level schemas in a single database, but does not allow multiple databases in the yaml file.  I don't think there is any reason we couldn't support multiple databases from the same connection, this was just an arbitrary design choice.

Assuming we want stick with one database per connection (probably simpler for now), we could still allow fully-qualified object names including database name, by providing some way for the caller or metadata to specify the database name.

One way to do this would be to augment the metadata to allow a top-level property for database name.  For example, `"database_name": "my_database"`.  This is somewhat awkward, since the other components of the name are part of the hierarchy of the YAML, but I think this is OK.

Another way to do this would be to allow the caller to pass in the database name to the connection's `NameCompare`, similar to how schema search path can be passed in now.  In this design, the database name (if present in the fully-qualified object name) would have to match the one passed in to the connection.  If no database name is passed in to the connection, we could choose to either throw an error, or silently accept it.

I'm leaning toward just adding a property to the root of metadata.  Thoughts?

We could also consider supporting multiple databases.  This would not be possible to do by extending the current metadata format, since schemas are currently at the top level.  We could take a breaking change to the metadata format to push database to top level.  The compatibility issues could be managed by introducing a version number to the metadata file format.  This is still a bit cumbersome, because it would end up requiring people to create an empty top-level for cases where the curator doesn't want to specify a database name, similar to how people sometimes currently use empty quotes as the schema name.  It might be worth it, though, and we could perhaps try to rationalize the metadata format to support more flexibly if people want to omit database name or schema name.  For example, we could use some heuristic whereby tables can exist at top level, and the metadata parser determines ancestor length by working up from the column definitions.

[...]

Issue #455, opened 2022-03-13
Support special cases in the grammar for `qualifiedTableName`


oskarkocol:
I run into a couple edge cases related to the identifier following the `FROM` statement. I came across these when spiking support for BigQuery engine so they don't really affect current users. The edge cases may have to be addressed with changes to code outside of the engine specific classes inheriting from `SqlReader` defined [here](https://github.com/opendp/smartnoise-sdk/tree/main/sql/snsql/sql/reader).

### Using table identifiers with a hyphen (`-`)
PostgreSQL and BigQuery engines support table names that contain as long as they are escaped. This definition of the `ESCAPED_IDENTIFIER` does not include hyphens and a statement like ``FROM `data-base`.`schema`.`table` ``
https://github.com/opendp/smartnoise-sdk/blob/186b81721c0f5c9ef9028e56a5ac142e9b362602/sql/snsql/sql/parser/SqlSmall.g4#L471-L476
**Possible solution:** probably something as simple as adding hyphen to the match. For example:
```
    | '"' (LETTER | DIGIT | '_' | ' ' | '-')*? '"'
```

### Escaping the entire table identifier
BigQuery grammar allows escaping the entire table identifier. For example `` `database.schema.table` `` in place of `` `database`.`schema`.`table` ``. This way of escaping is not supported by the parser. I assume that it has to do with the way how identifiers are defined and how later they are reused by: 
https://github.com/opendp/smartnoise-sdk/blob/186b81721c0f5c9ef9028e56a5ac142e9b362602/sql/snsql/sql/parser/SqlSmall.g4#L457-L459

**Possible solution:** I don't see a clear solution and would love to hear your input. My intuition would be to introduce a table specific identifier that would handle different ways to escape the tables ([credit](https://github.com/leonchen83/antlr4-mysql/blob/c75dd9eaf6778604c0ae188ebac0027f47b955e9/MySQL.g4#L1331-L1333)).
```
qualifiedTableName : TABLE_IDENTIFIER

TABLE_IDENTIFIER
    : ((IDENT '.')? IDENT '.')? IDENT
    | '"' TABLE_IDENTIFIER '"'
    | '`' TABLE_IDENTIFIER '`'
    ;
```
But I expect that this would introduce complexity down the dependency that would have to be resolved (`QN2`, `QN3` and `TABLE_IDENTIFIER` match both table and column identifiers). If there is a more succinct way to go about it, I'd love to hear about it.

[...]

PR #458, opened 2022-03-25
Allow hyphen in names, allow three-part names

joshua-oss:


joshua-oss:
Addresses #455, #456, #457

PR #459, opened 2022-03-27
Add support for BigQuery

oskarkocol:
This PR is a follow up on https://github.com/opendp/opendp/discussions/430 and is still a WIP.

# Intention
This PR aims to introduce:
* `BigQueryReader` allowing querying data sets hosted on BigQuery using smartnoise-sdk
* Github Actions workflow for CI that tests the BigQuery functionality. It uses Terraform to create the ephemeral cloud environment.
* Cover gaps in documentation relevant to the BigQuery support

# How to use
## The SDK
The user can either pass the GCP credentials to create a client or pass the client
```python
from snsql import *
from google.cloud import bigquery
PROJECT_ID = "<insert_project_id>"
client = bigquery.Client(project=PROJECT_ID)

# define metadata either as a .yml file or a dictionary
# metadata = {}
privacy = Privacy(epsilon=0.1, delta=1/10000)
reader = from_connection(client, metadata=metadata, privacy=privacy, engine="bigquery")
result = reader.execute("SELECT avg(income) as avg_income FROM `pums`.`pums`")
```
## CI checks
1. Create a service account with BigQuery Admin and Storage Admin roles
2. Use the `.secrets-demo` file to create `.secrets` file and fill out the fields that will be used to populate environment variables in the CI environment. GCP Credentials have to be contained in one line (remove new line characters `\n` from the default `.json` file provided by GCP)
3. Run the following:
```bash
act pull_request -W .github/workflows/bigquery.yml --secret-file "sql/.secrets"
```

# Todo's
- [x] Address the `400 Unrecognized name: pid` error in the CI checks (https://github.com/opendp/smartnoise-sdk/pull/459/commits/c9f2b6828c51693c0cf589e7529479e9930cbf64)
- [x] Address the `400 No matching signature for operator = for argument types: INT64, STRING` error in the CI checks (https://github.com/opendp/smartnoise-sdk/pull/459/commits/c9f2b6828c51693c0cf589e7529479e9930cbf64)
- [ ] ~Expand support for GCP credentials to be passed as a path to the credentials file (currently, it's only via an environment variable)~ (this is not necessary in context of tests running in a container)
- [x] Expand documentation (https://github.com/opendp/smartnoise-sdk/pull/459/commits/160b8da029d0235d333662b3c26ac58df4fdf7f6)
- [x] Review [dependencies using poetry](https://github.com/opendp/smartnoise-sdk/blob/c361eed885ac1cd39b8515bcc3cac0de4edf6f16/contributing.rst#managing-dependencies) (All set!)
- [x] Move requirements.txt under `tests/setup/bigquery` (https://github.com/opendp/smartnoise-sdk/pull/459/commits/4a134f12a9566bfbf92af3ee7b02f635315e9209)
- [x] Pull the name of the CI bucket from the `.secrets` (https://github.com/opendp/smartnoise-sdk/pull/459/commits/e3ff02daf8f1ae85a79ccd6cb2dc15b15ddf9268)

More info in the comments


oskarkocol:
# Open Questions
#### Tests querying `PUMS.PUMS` expecting `pid` column
@joshua-oss I would love hear your thoughts on addressing an error that I encounter in the CI checks. It looks related to how the tests are designed and I didn't find a good workaround. The error is to the fact that some tests use `PUMS_pid` metadata, but query `PUMS.PUMS` database (which does not contain the `pid` columns). For example, `test_grouping_clamp.py` queries:
https://github.com/opendp/smartnoise-sdk/blob/c361eed885ac1cd39b8515bcc3cac0de4edf6f16/sql/tests/query/test_grouping_clamp.py#L12

Which gets rewritten into:
```sql
SELECT
  COUNT ( DISTINCT pid ) AS keycount,
  SUM ( income ) AS sum_income,
  COUNT ( income ) AS count_income 
FROM (
  SELECT *,
  ROW_NUMBER ( ) OVER ( PARTITION BY pid ORDER BY RANDOM ( ) ) AS row_num 
  FROM (
    SELECT pid AS pid,
    CASE WHEN income < 0 THEN 0 WHEN income > 100 THEN 100 ELSE  income END AS income
    FROM PUMS.PUMS ) 
  AS clamped ) 
AS per_key_random
WHERE per_key_random.row_num <= 1
```
`PUMS.PUMS` does not contain `pid` column which results in an error:
` google.api_core.exceptions.BadRequest: 400 Unrecognized name: pid at [1:187]`

I realize that [tests/README.md](https://github.com/opendp/smartnoise-sdk/blob/c361eed885ac1cd39b8515bcc3cac0de4edf6f16/sql/tests/README.md#test-against-multiple-engines) explicitly mentions:
```
Note that we select from PUMS.PUMS in the PUMS_pid database. 
This is because the schemas are identical, apart from a hidden primary key. 
All queries that work against one should work against the other.
```
But I find it puzzling in the light that the rewriter uses the `pid` column.
* Should the `PUMS.PUMS` table on BigQuery be loaded from `PUMS_pid.csv`?
* Is there a different way to address it? Any suggestions would be super helpful 👍 

Affected tests:
```
 =========================== short test summary info ============================
| FAILED tests/accuracy/test_accuracy.py::TestExecution::test_no_accuracy - goo...
| FAILED tests/accuracy/test_accuracy.py::TestExecution::test_accuracy - google...
| FAILED tests/engines/test_db_count.py::TestDbCounts::test_db_counts - google....
| FAILED tests/engines/test_db_count.py::TestDbCounts::test_db_counts_star - go...
| FAILED tests/engines/test_db_count.py::TestDbCounts::test_db_counts_no_max_ids
| FAILED tests/engines/test_db_count.py::TestDbCounts::test_db_counts_distinct_pid
| FAILED tests/engines/test_db_count.py::TestDbCounts::test_count_null_impute
| FAILED tests/query/test_grouping_clamp.py::TestGroupingClamp::test_clamp_on
| FAILED tests/query/test_grouping_clamp.py::TestGroupingClamp::test_clamp_off
| FAILED tests/query/test_thresholds.py::TestQueryThresholds::test_yes_tau_laplace
| FAILED tests/query/test_thresholds.py::TestQueryThresholds::test_yes_tau_gauss
| FAILED tests/query/test_thresholds.py::TestQueryThresholds::test_no_tau - goo...
| FAILED tests/query/test_thresholds.py::TestQueryThresholds::test_no_tau_noisy
| FAILED tests/query/test_top.py::TestTopAndLimit::test_order_limit - google.ap...
| FAILED tests/query/test_top.py::TestTopAndLimit::test_no_order_limit - google...
| FAILED tests/query/test_top.py::TestTopAndLimit::test_queries - google.api_co...
| ======= 16 failed, 241 passed, 3 skipped, 1 warning in 172.88s (0:02:52) =======
```
#### Mixed type of `educ` column
When loading datasets to BigQuery, one has to specify the type of each column. Some tests use `educ` as a string:
https://github.com/opendp/smartnoise-sdk/blob/c361eed885ac1cd39b8515bcc3cac0de4edf6f16/sql/tests/query/test_thresholds.py#L137
and some as an integer:
https://github.com/opendp/smartnoise-sdk/blob/c361eed885ac1cd39b8515bcc3cac0de4edf6f16/sql/tests/query/test_thresholds.py#L72

This results in the BigQuery specific error:
`google.api_core.exceptions.BadRequest: 400 No matching signature for operator = for argument types: INT64, STRING. Supported signature: ANY = ANY at`

What's the right way to address it? Is the `overrides={'censor_dims': False}` meant to address this? Would it need an extra case specific to BigQuery?
https://github.com/opendp/smartnoise-sdk/blob/c361eed885ac1cd39b8515bcc3cac0de4edf6f16/sql/tests/query/test_thresholds.py#L131

joshua-oss:
This is great progress!

The test databases, `PUMS`, `PUMS_pid`, `PUMS_dup`, and `PUMS_null` are supposed to appear identical to the analyst.  You can see the example in the setup script for Postgres [1], we make 4 different databases, and create a copy of PUMS.PUMS in each one, loading from the respective CSVs.  Then the test fixture can return any one of the databases, and a query that works in one should work in the others.  The rationale for these variants is that we were encountering bugs where something would appear to be fixed, but would fail in some untested combination of database setup, such as failing when no `private_id` was present, or failing when NULLs were present.  So our default testing pattern is to have each query run against all 4 settings.

The test fixture is designed to cache connections to multiple databases (each one having a PUMS.PUMS table), so it can feed out the appropriate connection to the test runner.  This works well for typical database engines, since databases are distinct objects, and each connection is typically set to a specific database.  That is, the connection will be tied to a database, and all SQL statements issued against the connection will be applied to that database, without the SQL statements needing to specify the database.  If BigQuery can support that abstraction approach, that would be ideal.  If not, we can figure out a workaround.

Note that PR #458 adds support for database names as part of the query object, so we could perhaps workaround by allowing automatic rewriting to add database specifiers in the query (for test harness only).  However, it would be preferable to have database tied to connection, if at all possible.

The test databases have some differences in underlying data types, to ensure that we are testing the inbound conversions in the private reader, comparison operators in postprocessing (such as `HAVING` or `ORDER BY`), and type coercion in the engine.  For `educ`, since we typically treat it as categorical, we sometimes use int and sometimes use string (these are two of the 3 most common categorical types).  Similarly, `sex` could be a categorical of type boolean, string, or int.   It looks like BigQuery's type coercion rules [2] don't allow comparison of int and string, no doubt for good reasons.  In cases like these, we should just skip that test when the engine/metadata combination isn't supported.  You can use the skipping logic in line 134 of that test.

Some of these could be converted to use integer comparison (e.g. `WHERE educ = 8`), and this will still work regardless of what's specified in metadata, due to type coercion in the API.  But we want to keep at least a few that exercise the type coercion in the engine.

[1] https://github.com/opendp/smartnoise-sdk/blob/main/sql/tests/setup/postgres/PUMS/pums.sql
[2] https://cloud.google.com/bigquery/docs/reference/standard-sql/conversion_rules

oskarkocol:
Thanks for the explanation. This really clarifies my questions about the tests. At the moment, the datasets in the GCP are loaded into the same project and dataset. As a result, we have:
* `project_id.PUMS.PUMS` loaded from `pums.csv`
* `project_id.PUMS.PUMS_DUP` loaded from `pums_dup.csv`
* `project_id.PUMS.PUMS_NULL` loaded from `pums_null.csv`
* ...

But, it looks like the way that the Postgres tests are set up, would be equivalent to setting up a separate project for each dataset. Namely:
* `project_id_pums.PUMS.PUMS` loaded from `pums.csv`
* `project_id_dup.PUMS.PUMS` loaded from `pums_dup.csv`
* `project_id_null.PUMS.PUMS` loaded from `pums_null.csv`
* ...

> Note that PR https://github.com/opendp/smartnoise-sdk/pull/458 adds support for database names as part of the query object, so we could perhaps workaround by allowing automatic rewriting to add database specifiers in the query (for test harness only). However, it would be preferable to have database tied to connection, if at all possible.

In context of BigQuery, setting up a separate project for each test dataset sounds like an overkill (it'd require separate credentials as far as I know). 

I think that a good middle ground could be based on:
1. removing schema qualifiers (i.e `FROM PUMS.PUMS` would turn into `FROM PUMS`) from tests like [this one](https://github.com/opendp/smartnoise-sdk/blob/c361eed885ac1cd39b8515bcc3cac0de4edf6f16/sql/tests/query/test_thresholds.py#L96))
2. For BigQuery, loading each dataset with the same table name but a different schema. So for example:
* `project_id.PUMS.PUMS` loaded from `pums.csv`
* `project_id.PUMS_NULL.PUMS` loaded from `pums_null.csv`
* `project_id.PUMS_DUP.PUMS` loaded from `pums_dup.csv`
* ...
3. For BigQuery, associating tests with a schema (I believe that the Postgres tests could stay untouched). We could add a utility function similar to [_update_connection_string()](https://github.com/opendp/smartnoise-sdk/blob/500fee56e892ed8830657a7d9c8f0dee5e379bc1/sql/snsql/sql/reader/postgres.py#L54-L66):

```python
# for PUMS_DUP loaded from pums_dup.csv
self.database = 'PUMS_DUP'
config = bigquery.job.QueryJobConfig(default_dataset = '{}.{}'.format(PROJECT_ID, self.database))
client = bigquery.Client(project=PROJECT_ID, default_query_job_config = job_config)
```

As a result, each a query like:
```sql
SELECT * FROM PUMS LIMIT 100;
```
Would be considered by BigQuery to be [1]:
```sql
SELECT * FROM `PROJECT_ID.PUMS_DUP.PUMS` LIMIT 100;
```

This way, we'd support tests for BigQuery while leaving the existing code (for Postgres, MySQL) untouched. Any thoughts on this approach? Any risks that I didn't take into account?

[1] https://googleapis.dev/python/bigquery/latest/generated/google.cloud.bigquery.job.QueryJobConfig.html#google.cloud.bigquery.job.QueryJobConfig.default_dataset

joshua-oss:
> In context of BigQuery, setting up a separate project for each test dataset sounds like an overkill (it'd require separate credentials as far as I know).
> 
> I think that a good middle ground could be based on:
> 
> 1. removing schema qualifiers (i.e `FROM PUMS.PUMS` would turn into `FROM PUMS`) from tests like [this one]

OK, it looks like we need a general way for different engines to map the test tables.  Since this is only required for running unit/CI tests that share queries across different engines, it seems best to do it in a way that doesn't require changes to the SqlReader implementations.

I'm hesitant to switch all queries to use bare (no schema) table names, since different engines have different semantics for that.  For example, Postgres will match under `public`, and SQL Server will match under `dbo`.  Plus, that path isn't well-tested at the moment.

One option is to extend the semantics of `connections-unit` to be more flexible.  At the moment, the keys under `databases` each represent a single test table designed to be used for CI tests.  The value for each key represents the actual connection that will be used.  For example, in Postgres:

```
  databases:
    PUMS: pums
    PUMS_large: pums
```

The current semantics of `connections-unit` means that `PUMS` and `PUMS_Large` will coexist in the same database on the same connection, with different table names.  The fact that the table names are same or different is implicit and cannot be overridden, likewise with the schema.  We could instead use some convention like:

```
  databases:
    PUMS_dup:
      connection: pums
      database: pums_dup_tests
      schema: pums_dup
      table: pums
      format: "{schema}.{table}"
```

Since `connections-unit` is different for different engines, different engines could use different groupings of database and schema, as well as different naming conventions.  In the example above, a unit test written as `SELECT COUNT(*) FROM PUMS.PUMS` would actually run as `SELECT COUNT(*) FROM pums_dup.pums` when run against this particular engine.  If an engine needs a three-part name, that would be specified in the `format` property on the YAML.

This would require only a change to how the DbCollection test fixture works, and would also require interception of `execute()` calls to factory-supplied private readers to rewrite table names in the AST.  I can make both changes in the next day or two if it looks like this pattern would work for your tests.

oskarkocol:
> I'm hesitant to switch all queries to use bare (no schema) table names, since different engines have different semantics for that. For example, Postgres will match under public, and SQL Server will match under dbo. Plus, that path isn't well-tested at the moment.

Hey! This totally makes sense. I looked around more and from my understanding, we're looking at a BigQuery specific limitation. Supporting separate databases just as it's implemented now, wouldn't be an issue with Snowflake or Redshift.

> The current semantics of connections-unit means that PUMS and PUMS_Large will coexist in the same database on the same connection, with different table names. The fact that the table names are same or different is implicit and cannot be overridden, likewise with the schema.

This sounds like a really smart way to go about it and it aligns with a broader approach of 3-part names that were introduced in https://github.com/opendp/smartnoise-sdk/pull/458

> This would require only a change to how the DbCollection test fixture works, and would also require interception of execute() calls to factory-supplied private readers to rewrite table names in the AST. I can make both changes in the next day or two if it looks like this pattern would work for your tests.

Would you be able to bring this change over especially that it sounds like a small piece of work for you? I think that I'd need a little bit more discussion if I was to bring it myself 😅 (and probably it'd be better to keep it out of this PR because it's starting to grow large).

@joshua-oss On a separate note - I'm introducing here CI checks built on Github Actions. They require setting up GCP credentials on Github Secrets that would be used to create ephemeral testing environments. Is OpenDP happy to set up and maintain the creds as an organization? If not, I'll try to think about an alternative.

There is no need to set them up right away because I can test if things work as expected on my fork. But I imagine, it's something that would have to be sorted out before this PR could be merged.

joshua-oss:
OK, I have made PR #466 to allow `connections-unit` to specify a table name to be used when querying that test dataset with the test fixture.  This PR will make sure the metadata uses the overridden table name, and will replace the table name in the query.  You can see an example in the `connections-unit` under `tests/setup/postgres/PUMS` in that PR.

Since it's a breaking change to `connections-unit`, I changed the syntax from `databases` to `datasets` to allow the code to distinguish if an older version of the file format is being used so it can warn the developer to update.

I didn't implement as a "format" string, since I couldn't think of any scenario where the target table name would need to be parameterized like that.  The target table name can just be spelled out in the `table` property.  

Note that the 3-part names support doesn't support escape around the full three-part name (e.g. "PROJECT_ID.PUMS_DUP.PUMS"), since I verified that this would break in postgres and sql server.  That is, the other engines treat this as an attempt to escape the table part of the name, rather than escaping all 3 parts. 
 However, I am assuming that's OK, because I assume BigQuery supports escaping like: "PROJECT_ID"."PUMS_DUP"."PUMS".  Let me know if that's not a safe assumption.
 
 I'll merge this into main in the next couple of days, and then we can merge main back into your PR.  I refactored the test fixture a little bit to compartmentalize code for different engines, so we will need to update the fixture to support BigQuery; this is probably just boilerplate.
 
We will be able to host the CI job using a secret stored in our organization.  I think I can create a GCP account and generate the necessary secret, so then we will handle any compute charges, which I assume should be pretty small. 

One other thing I forgot to mention; we can move the pip requirements that are specific to BigQuery, into the requirements.txt under `tests/setup/bigquery`.  We do this for all of the other engines, and the distinct CI jobs will call `pip install -r` for just the engine that is being tested.  The assumption is that someone who is querying a specific engine will already have a way to connect to that engine, and they will just pass the connection in to `PrivateReader.from_connection()`.

joshua-oss:
Woo hoo!  I was able to get all unit tests passing!

I pulled your PR into a local branch for testing:
https://github.com/opendp/smartnoise-sdk/tree/test_bigquery

Then I merged in the branch from PR #466.  A few more changes were needed to make everything compatible with the new test fixture design.  Then a few other changes:

* `BigQueryReader` switches `RANDOM` to `rand`.
* Switch schema to use integer for `educ` on `PUMS` to allow threshold tests to pass
* Preprocess PUMS_large to remove double-quotes so PersonID can be parsed as integer by BigQuery
* Preprocess `connections-unit.yml` to use `GOOGLE_PROJECT_ID` in the table names

I didn't see a way to make the changes directly in your PR, but maybe you can merge the `test_bigquery` branch back into your PR, and review to make sure the changes make sense?

Assuming everything looks OK to you, there are a couple of other things to take care of:

1. The storage bucket name apparently needs to be globally unique.  I made one under my account and search/replaced everywhere that looked relevant.  You'll need to change these back to point to your storage bucket.  If it's possible to make this dynamic based on secrets, that would be ideal.
2. I'm doing simple string replace for the `RAND` switch.  Assuming everything else looks OK, I'll go in and fix that to use the AST, after I approve and merge the pull request
3.  Anything else?

oskarkocol:
What an exciting progress! I'm back from holidays and I'm going to start putting everything together this week.

At this point, the TODO list seems clear 🤞 Thank for your input @joshua-oss 🙇

oskarkocol (commented):
Hey @joshua-oss this PR is ready for a review 🎉 I would love to hear your feedback. I applied all the changes that we talked about and I tested the new CI tests with the ephemeral environment on my fork - things look green 💚  See https://github.com/oskarkocol/smartnoise-sdk/pull/3 for more

image

Other than the comments below, I have one open question:
* Is there any documentation other than the READMEs that I touched and docs that you've updated, that should get updated? Is there a need to recompile the docs?

joshua-oss (approved):
Looks good!  This is a big contribution; thanks for working through this!

oskarkocol:
My pleasure! Thanks for https://github.com/opendp/smartnoise-sdk/pull/458 that unblocked this PR! Happy to help with setting up the Github Workflow 👍

PR #463, opened 2022-04-05
add support for continuous columns for GAN based models

AprilXiaoyanLiu:
Previous issue: we stop to support continuous columns because of potential privacy leakage when transforming the continuous variables in GAN models
Solution: I added three different transformers for users to choose to support continuous columns. 
1. BaseTransformer in data_transformer.py:  no change for continuous column data, one-hot for categorical columns:  If your continuous column data is too large, may have some gradient explosion issue
2. DPMinMaxTransformer in dpminmax_transformer.py:   implement differential private minmax scaler to transform continuous columns to the range of (-1, 1). Need specify continuous_columns_lower_upper 
3. DPSSTransformer in dpss_transformer.py: apply differential private standard scaler to transform continuous columns


joshua-oss (approved):
Thanks for fixing this!

PR #469, opened 2022-05-15
MST update

lurosenb:
Cleaned up MST file a bit, added an MST sample that details creating the domain file


joshua-oss (approved):
Looks good; thanks!

Issue #472, opened 2022-06-14
MWEM laplace noise


gergelyacs:
in mwem.py: 
```
lap = self._laplace(
                    (2 * self.iterations * len(self.histograms)) / (self.epsilon * len(dimensions))
                )
```
why is len(dimensions) in the denominator?
I have a 4D histogram (dataset with 4 attributes), where the sensitivity of a single query is still 1 (or 2), then you apply composition over all iterations (self.iterations), and use the exponential mechanism (hence factor 2 in the nominator). But why do we have to divide by the number of dimensions (in my case, 4)?

joshua-oss:
Hmm, I don't remember why we did this.  It doesn't seem to be necessary per the paper [2].  @lurosenb, do you recall why we reduce the budget by a factor of the number of dimensions in the line at [1]?

[1] https://github.com/opendp/smartnoise-sdk/blob/main/synth/snsynth/mwem.py#L266
[2] https://www.cs.huji.ac.il/~katrina//papers/mwem-nips.pdf

joshua-oss:
Update on this; I am pretty sure you are correct.  We will be deploying a method to test the empirical privacy of our synthesizers, and will fix this when we can show empirically with that tool that we can safely less noise and stay within the overall privacy budget.

lurosenb:
Apologies for the delay on this! 

You are correct, I believe this is an artifact of a faulty sensitivity calculation for linear range queries. I'll hot fix this immediately - thanks for pointing it out!

lurosenb:
Please approve https://github.com/opendp/smartnoise-sdk/pull/474 @joshua-oss

joshua-oss:
Thanks for the fix, @lurosenb!  Approved, merged, and released to PyPi

PR #474, opened 2022-06-28
Fix Issue #472

lurosenb:
Thanks @gergelyacs !


joshua-oss (approved):
Looks good; thanks for the fix!

Issue #473, opened 2022-06-20
differential privacy of numpy.histogramdd


gergelyacs:
in mwem.py, `np.histogramdd` is called from `def _histogram_from_data_attributes`.
As it uses the original data to create bins, shouldn't the identification of bin edges be randomized too in order to guarantee Differential Privacy?

joshua-oss:
Thanks for flagging this.  Our intention is for MWEM to only ever receive categorical columns, with the categories provided externally.  The way the code is currently written, it's not actually trying to discretize continuous values; it's just treating all columns as if they are integer-coded categorical and giving each category its own bin.  IOW, the assumption is that the data are categorical, with categories being like (0, 1, 2, ...).  This approach creates risk in at least two ways.  First, if the caller has some external category encoding for integers where not all integers map to categories (e.g. the data dictionary specifies categories [1,2, 4]) the histogram we create will have spurious categories.  Second, if the caller doesn't realize these are meant to be categories, it could result in some leakage and poor utility.

We are intending to fix both of these in the next month by refactoring the way we pass the data to the synthesizers.  You can see the proposal here, please let us know if you have any feedback: https://github.com/opendp/smartnoise-sdk/issues/467

I will keep this issue open until that PR is merged

joshua-oss:
MWEM now uses a `LabelTransformer` for all categorical columns by default, so all categorical columns are stored in integer indices starting at zero.  Continuous values are binned externally by the TableTransformer, using privacy budget to infer approximate bounds if bounds are not provided.

Issue #467, opened 2022-04-26
Discussion: Reversible Data Transforms


joshua-oss:
This issue is for discussion of the API for data transforms.

# Overview

PROBLEM STATEMENT: Various differentially private data processing routines have different expectations about the format of input data, so it's often necessary to preprocess and postprocess.  For example, a GAN-based synthesizer expects categorical values to be one-hot encoded, while MWEW-style synthesizers need all values to be discretized into integer values.  Pushing these requirements to the analyst gives the analyst greatest control, but can lead to privacy leaks, since many common preprocessing approaches use summary information about the data.

GOAL: Provide a set of *safe* reversible data transforms that make it easy for analysts to adapt data for the processing algorithms in SmartNoise, particularly for the synthesizers.  Data transforms should be able to run independently; with minimal or no coupling to synthesizers.  Transforms should keep track of all privacy spend, to allow efficient accounting.

PROPOSAL: Adopt the basic interface of RDT [1] for column-value transforms.  Specifically, all column-transforms have `fit()`, `transform()`, and `reverse_transform()` methods.  Allow both chaining of column transformers on a single column, as well as aggregation of multiple column transformers to transform a table (e.g. a `TableTransformer` patterned after SDT's `HyperTransformer`)

# Base Column Transformer

Will use the name `ColumnTransformer` here for discussion purposes; other names may be better.

```python
class ColumnTransformer:
   def fit(self, data)
   def transform(self, data)
   def reverse_transform(self, data)
```

## Constructor

Instances of concrete `ColumnTransformer` implementations may have instance-level state that is used to keep track of transformation parameters or privacy budget.  Instance-level state can be provided in the constructor, or may be changed in the `fit()` method.  For example, a transformer that uses public bounds to scale the value could use syntax like:

```python
transform = MinMaxScaler(min=0, max=100)
```

while the same transformer could spend some privacy budget to infer the bounds, like:

```python
transform = MinMaxScaler()
transform.fit(pums, 'age', epsilon=0.1)
```

The `transform` and `reverse_transform` functions do not change any instance state.  The output schema of `transform` will be the input schema of `reverse_transform` and vice-versa, and these two schemas will not usually be the same.  For example, a `BinTransform` may take floats and return sequential integers.  A `OneHot` transformer will take sequential integers and return multiple bit columns.

[...]

Issue #465, opened 2022-04-09
Synthesizer Factory and Interface


joshua-oss:
Thread for synth factory design brainstorm and discussion.

### Goals

Current design requires callers to know concrete synthesizer class name, import appropriate classes, and in some cases call with a wrapper.  We would like to simplify allow any synthesizer to be created from the same basic imports.  Design should support easy addition of new synthesizers.

### Creation

Some options described by idealized code samples:

1. Synthesizer with default hyperparameters

```python
from snsynth import Synthesizer

synth = Synthesizer.create("mwem")
synth.fit(df)
data = synth.sample(50)
```

2. Override a hyperparameter

```python
from snsynth import Synthesizer

synth = Synthesizer.create("patectgan", batch_size=50)
synth.fit(df)
data = synth.sample(50)
```

3. We could allow multiple factory entries for the same concrete synthesizer, using different hyperparameters.  For example, a PATECTGAN instance with hyperparameters optimized for smaller datasets might be created like:

```python
from snsynth import Synthesizer

synth = Synthesizer.create("patectgan_small")
synth.fit(df)
data = synth.sample(50)
```

A drawback of using strings to switch is that callers will need to refer to the documentation to know what synthesizer configurations are available.  This could be mitigated by providing introspection on the factory allowing enumeration and description of the synthesizers:

```python

from snsynth import Synthesizer

for synth_name, description in Synthesizer.synthesizers:
   print(f"{synth_name}: {description})
```

The synthesizer list would be visible from the debug watch in the IDE, but would not populate options automatically via intellisense.

Alternately, we could switch based on an enum:

```python
from snsynth import Synthesizer, Synth

synth = Synthesizer.create(Synth.mwem)
synth.fit(df)
data = synth.sample(50)
```

The allowed enum values will populate in the IDE automatically, self-documenting via intellisense, and reducing the odds that the caller will ask for a synthesizer that doesn't exist.  However, this adds an extra import, and feels repetitive in the code.

Is there a better way to key the factory??

### Interface

In the factory examples above, the factory method is a static method on the base class that also defines the interface for all synthesizers.  All concrete instances of the `Synthesizer` base class should have at least two methods and a property:

* `fit()` -- learns the distribution of a non-private dataset
* `sample()` -- draws synthesized records based on the learned distribution
[...]

Issue #445, opened 2022-01-24
Not working on mac M1


s-inder-preet:
[smartnoise_error.log](https://github.com/opendp/smartnoise-sdk/files/7925255/smartnoise_error.log)
Getting this error while installing smartnoise-synth on a Macbook pro with Apple silicon. 

Used the commmand:
`python3 -m pip install smartnoise-synth`

joshua-oss:
Unfortunately, a few of the packages we depend on still do not have M1 builds.  I use an M1 MacBook, however, and smartnoise-synth works well in using 64-bit python, such as "Miniconda3 MacOSX 64-bit pkg".

joshua-oss:
Fixed in v0.3.0

PR #490, opened 2022-10-08
Synthesizer Factory, Transforms, and M1 Support

joshua-oss:


joshua-oss:
Addresses #445
Addressed #467
Addresses #465
Addresses #473
Addresses #433

PR #499, opened 2022-10-27
Fix broken link in sql/docs/README.md

TedTed:


joshua-oss (approved):
Thank you for catching this!

PR #500, opened 2022-10-28
Update to use OpenDP v0.6

joshua-oss:


Shoeboxam:
Thanks for updating! Hoping to have more api stability moving forward. We pushed the module renames through because we're stabilizing the proof versioning scheme.

Shoeboxam (approved):

Issue #505, opened 2022-11-20
Add anonymization transformer


neat-web:
### User story
As a user of the `smartnoise-synth` library, I frequently want to anonymize personally identifiable information (PII) in the data.

### Detailed description
To prevent leakage of sensitive information PII is anonymized by an analyst when preprocessing the data. Generating fake data is a solution because it discards existing values and instead uses entirely new ones. The most popular library for this is [Faker](https://pypi.org/project/Faker).

### Proposed solution
One possibility would be to follow the idea of the Synthetic Data Vault (SDV) [implementation](https://github.com/sdv-dev/SDV/discussions/667). It integrates `Faker` and allows a user to pass an `anonymize_fields` keyword argument. This maps a column to a corresponding provider. Even though that is quite user-friendly, it restricts the usage of more complex generation ideas, e.g. providing a custom implementation.

Therefore, I suggest not to integrate `Faker` directly, but instead let the user pass a function for generating fake data. Since data transforms can be provided in a `TableTransformer` object, I suggest implementing a new `ColumnTransformer` class for anonymization.

Feel free to make suggestions. I will implement the functionality soon.

joshua-oss:
That's a great idea.  I think it's OK to take a dependency on faker for their built-ins, but I also like the idea of allowing overrides.  As you suggest, this should be possible by deriving `snsynth.transform.base.ColumnTransformer` .  Since it will be assuming the column is PII, it shouldn't be necessary to implement anything for `_fit`, just for `_transform` and `_inverse_transform`.  Presumably the `_transform` just scrubs the value, and `_inverse_transform` generates fake PII when reversing the transform.

The current `ColumnTransformer` design allows `_transform` to map from 1 to n columns, with `_inverse_transform` mapping back from n to 1.  I'm not entirely sure the best way to handle this with, for example, a single column of PII.  Conceptually, the whole PII column is dropped, but then we have nothing to invert back (we can't map 1-->0 and then 0<--1).  An easy choice would be for `_transform` on the anonymizing `ColumnTransformer` to emit a column of `None` that the `TableTransformer` can then ignore when feeding the synthesizers (and re-inject on reverse transform).  So, the anonymizing column transformer would only emit the column of `None` when operating standalone or in a `ChainTransformer`.

[...]

PR #507, opened 2022-11-23
Implement anonymization transformer

neat-web:
In this PR I implemented an `AnonymizationTransformer` as discussed in #505.

I also added a dependency for the `Faker` library and documented an example in the transformers documentation.

### Implementation details
As suggested, the existing values are discarded during transformation and new ones will be generated during `_inverse_transform()`.

The transformer supports `Faker`. Even the configuration of e.g. `locale` and custom providers can be done. Apart from these built-ins, a custom callable can be provided.

If someone wants to use the fakes for training of a synthesizer, the optional argument `fake_inbound=True` can be provided. Then the new values will be injected during transformation and are passed through on inverse. This also enables usage in a `ChainTransformer`.

### Proposal
We could raise an error if an `AnonymizationTransformer` is given in a chain without having `fake_inbound` set to true. I would refrain from setting the argument implicitly or discarding `None` values in general, and instead make the user aware. Otherwise, one could falsely assume that the synthesizer is not trained with fake values, when he actually is.

Even though the tests achieve 100% code and branch coverage, I think it would be a good idea to add tests in `TestTableTransform`.

### DIFF
--- a/synth/docs/source/transforms/index.rst
+++ b/synth/docs/source/transforms/index.rst
@@ -167,6 +167,35 @@ If this argument is not provided, the synthesizer will attempt to infer the most
     print(f"DPCTGAN inferred a onehot transformer with {dpctgan._transformer.output_width} columns")
 
 
+Anonymize personally identifiable information (PII)
+------------------------
+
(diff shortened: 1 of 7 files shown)


joshua-oss (approved):
This looks good.  I see you had to override the `transform` and `fit` base class methods.  I'll approve and merge this, and then add some tests with `TableTransformer`.

Issue #509, opened 2022-11-25
Error during preprocessing for a GAN-based synthesizer


neat-web:
### Subject of the issue
A `TypeError` is raised when preprocessing for a GAN-based synthesizer with a categorical column that contains just one value.

### Environment
* Windows 10, Python 3.8
* `smartnoise-synth` installed via pip (version 0.3.2)

### Steps to reproduce
I installed the latest `smartnoise-synth` version and tried to train a GAN-based synthesizer e.g. `PATECTGAN`. The data contains a categorical column with just one value. I used an inferred `TableTransformer` with the default one-hot style for preprocessing.

### Expected behaviour
Preprocessing works and the synthesizer can be trained.

### Actual behaviour
A `TypeError` is raised during preprocessing. The traceback is shown below:
```python
Traceback (most recent call last):
  ...
  File "...\venv\lib\site-packages\snsynth\pytorch\nn\patectgan.py", line 551, in fit
    self.train(data, transformer=transformer, categorical_columns=categorical_columns, ordinal_columns=ordinal_columns, continuous_columns=continuous_columns, preprocessor_eps=preprocessor_eps, nullable=nullable)
  File "...\venv\lib\site-packages\snsynth\pytorch\nn\patectgan.py", line 214, in train
    train_data = np.array([
  File "...\venv\lib\site-packages\snsynth\pytorch\nn\patectgan.py", line 215, in <listcomp>
    [float(x) if x is not None else 0.0 for x in row] for row in train_data
  File "...\venv\lib\site-packages\snsynth\pytorch\nn\patectgan.py", line 215, in <listcomp>
    [float(x) if x is not None else 0.0 for x in row] for row in train_data
TypeError: float() argument must be a string or a number, not 'tuple'
```

### Possible solution
The current implementation of the `OneHotEncoder` outputs a tuple, even if its `output_width` is one. Thus, the `TableTransformer` appends the tuple to the `out_row` list. This results in a `TypeError` later on during creation of the training data array.

neat-web:
The issue is fixed in version 0.3.3.

The discussion in the related PR concludes that there is a slight incompatibility with transformers from the `sklearn` library that is not relevant currently. Also, the sporadically failing CI tests might lead to a deprecation of poorly performing GAN based synthesizers in the future.

PR #510, opened 2022-11-25
Fix OneHotEncoder implementation

neat-web:
This PR adresses the error encountered in #509.

### Implementation details
I implemented a simple test for the described problem. Then I fixed the current `OneHotEncoder` implementation by returning the value directly instead of a tuple.

### DIFF
--- a/synth/snsynth/transform/onehot.py
+++ b/synth/snsynth/transform/onehot.py
@@ -27,6 +27,11 @@ class OneHotEncoder(ColumnTransformer):
     def _transform(self, val):
         if self.max < 0 or not self._fit_complete:
             raise ValueError("OneHotEncoder has not been fit yet.")
+        elif val < 0 or val > self.max:
+            raise ValueError(
+                f"Provided integer-label {val} is invalid."
+                " Please ensure that all inputs are 0-based and provided during data fit."
+            )
         elif self.max == 0:
             return 1
 
--- a/synth/snsynth/transform/onehot.py
+++ b/synth/snsynth/transform/onehot.py
@@ -27,6 +27,9 @@ class OneHotEncoder(ColumnTransformer):
     def _transform(self, val):
         if self.max < 0 or not self._fit_complete:
             raise ValueError("OneHotEncoder has not been fit yet.")
+        elif self.max == 0:
+            return 1
+
         bits = [0] * (self.max + 1)
         bits[val] = 1
         return tuple(bits)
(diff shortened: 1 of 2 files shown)


synth/snsynth/transform/onehot.py
```diff
@@ -27,6 +27,9 @@ def _clear_fit(self):
     def _transform(self, val):
         if self.max < 0 or not self._fit_complete:
             raise ValueError("OneHotEncoder has not been fit yet.")
+        elif self.max == 0:
+            return 1
```
joshua-oss, line 31: Can we throw an error here if self.max == 0 and val != 0?

neat-web: By the way: Your CI tests keep failing sporadically. The culprit seems to be `TestFactory.test_fit_with_data_frame()` if the sample of a synthesizer does not contain one row with `married == 0`. I can fix the test if you suggest what it should look like instead.

joshua-oss: This fix looks great; I'll merge and push a release.  Regarding the CI, I can ignore the failure for now, since it just appears to be catching a very poor fit for the non-CT DPGAN, and I haven't decided what to do about that.  The non-CT GAN-based synthesizers aren't very good, so maybe we deprecate them, or at least stop validating their outputs, or maybe do some more checking to see if they are better with continuous values or something?

joshua-oss: One other comment: the sklearn OneHotEncoder returns a list of lists even if there is only one category, and there are good reasons to do that.  But as you noticed, we break downstream if there are any tuples of width 1, since we assume that all columns can be nullable, which immediately turns them into a tuple, and so we use output_width == 1 as a way to tell that something should be a value rather than a tuple.  We could rework everything downstream to preserver closer compatibility with sklearn in this edge case, but I think your solution is better.

joshua-oss (approved):
Looks good; thanks for the fix!

synth/snsynth/transform/onehot.py
neat-web: Okay good. Seems like you just closed the PR without merging (by accident?).

> The non-CT GAN-based synthesizers aren't very good, so maybe we deprecate them, or at least stop validating their outputs, or maybe do some more checking to see if they are better with continuous values or something?

Since they do not provide good results for synthetic tabular data, I would second your deprecation proposal. Or at least make it clear to a user that the model is untested and might be an inadequate synthesizer choice.

> We could rework everything downstream to preserver closer compatibility with sklearn in this edge case, but I think your solution is better.

Thank you for providing your insight. I don't think there is a need to rework the internals, at least not currently. A user can use sklearn separately, without a `TableTransformer`.

PR #514, opened 2022-12-01
Makes postgres engine commit every time a query is made

FishmanL:
Since we only have read queries, not writes, it's safe to autocommit per transaction, and this prevents weird engine errors where one failed txn means all future txns with that PrivateReader will fail.


joshua-oss (approved):
Looks good, thanks for the fix!

PR #519, opened 2022-12-06
Default integers to use 64-bit depth

joshua-oss:


FishmanL (commented):
lgtm!  maybe have int vs bigint in the metadata?

Issue #517, opened 2022-12-05
Constrain inferred preprocessor


neat-web:
### User story
As a user of the `smartnoise-synth` library, I want to provide a variable amount of `ColumnTransformer` objects for preprocessing during `fit()`. Especially if the data contains many columns or the majority of the inferred transformers is OK, this can save time and repeated definitions.

### Detailed description
Currently, one either specifies the whole preprocessor (i.e. at least one transformer per column) or allows the synthesizer to infer the preprocessor completely. By specifying only certain `ColumnTransformer` objects, a user can constrain the inferred preprocessor to certain columns.

For my use case I would like to enforce that e.g. the "age" column uses an `AnonymizationTransformer` even when I am content with the other inferred transformers.

### Proposed solution
In my opinion, one could specify an optional parameter `fit(transformer_constraints=...)` which takes a dictionary that maps from column indices / names to a `ColumnTransformer` object. Assuming `transformer=None` is given, then all except the referenced columns will be inferred.

Another approach is to modify the `TableTransformer` class directly. Instead of assuming that the parameter `transformers` is solely a list, it could be similar to the described dictionary. If some columns are missing in the given `TableTransformer` object, they can be inferred automatically.

I will open a new branch and work on this once we agree on the specifics.

joshua-oss:
Good idea.  

Regarding the two proposals, the `SDGymBaseSynthesizer.fit()` method currently takes three lists (`ordinal_columns`, `categorical_columns`, and `continuous_columns`), which are passed through to the `TableTransformer` factory methods.  The idea is that creating a `TableTransformer` manually and passing to the synthesizer, or allowing the synthesizer to create a `TableTransformer` under the covers, will result in the same behavior.  Ideally, whatever is designed for the above proposal should have the same symmetry.

The current design was a holdover from our SDGym heritage, where there were only 2 or 3 column types allowed, and any column not specified in one of two types would be assumed to be the third type.  That was problematic in our setting, so we switched to requiring all columns to be explicitly listed, and we throw an error (rather than assuming) if a column is present in the training data but has an unknown type.  The inference code mitigates this somewhat, but the point is that the code requires all columns present to have a transformer, and only allows transformers to be generated from 3 types (ordinal, categorical, and continuous).

[...]

PR #522, opened 2022-12-11
Implement preprocessor constraints

neat-web:
In this PR I implemented the possibility to constrain the preprocessor according to #517.
Furthermore, I implemented a `DropTransformer` that enables a user to completely drop a column.

### Implementation details
I added the functionality on top of the legacy approach for SDGym. This means that a column can be referenced either in one of the three lists or as a constraint. I am in favor of maintaining the lists for compatibility reasons, at least until the next major release.

As proposed, a column referenced in the optional `constraints` dictionary is not inferred. There are multiple ways to specify a constraint, have a look at the `TableTransformer.create()` method documentation. A constraint might use privacy budget if e.g. a `MinMaxTransformer` is enforced without providing the bounds.

The `constraints` dictionary can be provided instead of a `TableTransformer` instance. This prevents a new parameter from being required. However, I have not touched the `special_types` dictionary yet. Maybe you can take care of that if you agree with the proposed approach?

Here is an example which showcases different usage scenarios:
```python
import pandas as pd
from snsynth import Synthesizer
from snsynth.transform import *

pums = pd.read_csv("PUMS.csv", index_col=None, usecols=["age", "income", "married"])

for synth_name in Synthesizer.list_synthesizers():
    print(f"synth: {synth_name}")

    # infer all columns except age
    synth = Synthesizer.create(synth_name, epsilon=1.0)
    synth.fit(pums, transformer={"age": "address"}, preprocessor_eps=0.5)
    print(synth.sample(10))

    # infer all columns except income
    synth = Synthesizer.create(synth_name, epsilon=1.0)
    synth.fit(pums, transformer={"income": AnonymizationTransformer(lambda: 123)}, preprocessor_eps=0.5)
    print(synth.sample(10))

    # don't infer any column type
    synth = Synthesizer.create(synth_name, epsilon=1.0)
    synth.fit(pums, transformer={"age": "ordinal", "income": "continuous", "married": "categorical"},
              preprocessor_eps=0.5)
    print(synth.sample(10))
```

### Proposal
A usage example in the data transformers documentation should be added.
At least some end-to-end test like in `TestFactory` would be good as well. I did not include the above example as a test because the runtime is quite long.

### DIFF
--- a/synth/snsynth/base.py
+++ b/synth/snsynth/base.py
@@ -146,16 +146,17 @@ class Synthesizer(SDGYMBaseSynthesizer):
         """
         return list(synth_map.keys())
     def _get_train_data(self, data, *ignore, style, transformer, categorical_columns, ordinal_columns, continuous_columns, nullable, preprocessor_eps):
-        if transformer is None:
+        if transformer is None or isinstance(transformer, dict):
             self._transformer = TableTransformer.create(data, style=style,
                 categorical_columns=categorical_columns,
                 continuous_columns=continuous_columns,
                 ordinal_columns=ordinal_columns,
-                nullable=nullable,)
+                nullable=nullable,
+                constraints=transformer)
         elif isinstance(transformer, TableTransformer):
             self._transformer = transformer
         else:
-            raise ValueError("transformer must be a TableTransformer object or None.  See the updated documentation.")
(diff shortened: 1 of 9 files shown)


joshua-oss (approved):
Looks great; thanks!

PR #530, opened 2023-02-10
Add AIM synth

aholovenko:
I would like to commit to creating a wrapper for AIM synthesizer from [private-pgm repo](https://github.com/ryan112358/private-pgm/blob/master/mechanisms/aim.py)

### DIFF
--- a/synth/README.md
+++ b/synth/README.md
@@ -11,6 +11,7 @@ Differentially private synthesizers for tabular data.  Package includes:
 * DP-CTGAN
 * PATE-CTGAN
 * PATE-GAN
+* AIM
 
 ## Installation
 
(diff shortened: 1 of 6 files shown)


joshua-oss:
Thanks for this contribution!  It will be great to have a good workload-aware synthesizer available.

One question: where is the import for HDMM (`from hdmm.matrix import Identity` in `aim.py`) coming from?  I assume it's from ektolo, but can't find a reference in either of Ryan's repos, and it doesn't install with `pip install git+https://github.com/ryan112358/private-pgm.git`.

lurosenb:
Note: we only need one version of cdp2adp - will remove the extra file in a follow up mini-PR

lurosenb (approved):
Thanks so much for the contribution Stacy!!!

Issue #560, opened 2023-05-12
Issues with Smartnoise SQL samples


oskarkocol:
Hey hey,

Inspired by the recent updates https://github.com/opendp/smartnoise-sdk/pull/558 to the documentation, I want to highlight couple of problems with the example notebooks in the `smartnoise-sql` [`samples`](https://github.com/opendp/smartnoise-sdk/tree/88038058f727fe9f87a1ab2683b1d08024a7327d/sql/samples) folder and offer a fix.

### Policy Laplace on Spark - [link](https://github.com/opendp/smartnoise-sdk/blob/main/sql/samples/Policy%20Laplace%20on%20Spark.ipynb)
1. The dataset doesn't exist anymore at this location, so this code will never work.
```python
import wget
zip_path = 'https://dp-test-datasets.s3.amazonaws.com/clean_askreddit.csv.zip'
wget.download(zip_path)
```
I imagine, we're still using the same dataset for tests:

https://github.com/opendp/smartnoise-sdk/blob/88038058f727fe9f87a1ab2683b1d08024a7327d/sql/tests/setup/dataloader/download_reddit.py#LL22C3-L22C3

2. Finding [`policy_laplace.py`](https://github.com/opendp/smartnoise-sdk/blob/main/sql/samples/policy_laplace.py) is not very intuitive unless the user has cloned the repo. Initially, I thought that `policy_laplace` is a part of `smartnoise-sdk`.

### SQL Queries with Differential Privacy - [link](https://github.com/opendp/smartnoise-sdk/blob/main/sql/samples/SQL%20Queries.ipynb)

1. This example is relevant to someone who has set up the development environment for `smartnoise-sdk` repo but not for a casual user of the package. Running the first cell requires following [these instructions](https://github.com/opendp/smartnoise-sdk/tree/main/datasets#downloading-datasets) beforehand:
```python
pums = pd.read_csv('../../datasets/PUMS.csv')
```

2. Similar issue when metadata in `.yaml` files is referenced.

### Synopsis - [link](https://github.com/opendp/smartnoise-sdk/blob/main/sql/samples/Synopsis.ipynb)
Similar issues as in `SQL Queries with Differential Privacy`.

## Proposed solution
- Update the references to test datasets
- Make the notebooks easy to run in the browser without the need to clone the repo
- Add direct links to free cloud notebook providers such as Google Colab or Kaggle

oskarkocol:
[...]

PR #561, opened 2023-05-12
Smartnoise SQL samples update

oskarkocol:
Addressing https://github.com/opendp/smartnoise-sdk/issues/560

1. Updating the reference to the datasets
2. Make the notebooks easy to run in the browser without the need to clone the repo
3. Adding links to Kaggle and Colab

image

### Open questions for the reviewer:
1. I added contents of https://github.com/opendp/smartnoise-sdk/blob/main/sql/samples/policy_laplace.py into the Laplace Notebook. Are we ok with deleting `policy_laplace.py`?
2. Do we want to have notebooks committed to the repo with or without the cell output? Similar examples maintained by Google do not have any output ([Vertex AI example](https://github.com/GoogleCloudPlatform/vertex-ai-samples/blob/main/notebooks/official/datasets/get_started_bq_datasets.ipynb)). This PR not include cell outputs. Current notebooks do contain cell outputs.


joshua-oss (approved):
This is fantastic; thanks for doing this!

We have had a few people run into issues with missing PUMS.csv when running code examples from the docs, so I will merge this PR and then push another change to include the files in Git LFS so that people with LFS installed will get the files automatically when cloning

PR #578, opened 2023-10-20
Adding details to README for new users

crangelsmith:
Today I was trying to run the examples in the README for the first time and found that some extra comments could have saved me some time (e.g where the `PUMS.csv` and `PUMS.yml` are found).

I've updated the READMEs with those comments with help from @mhauru.


joshua-oss (approved):
Thank you so much for these improvements to the documentation!

Issue #573, opened 2023-09-29
Code snippet in documentation does not work


TedTed:
I tried to run the code in the "Preprocessor hints" section of [this documentation page](https://docs.smartnoise.org/synth/index.html), and it fails for multiple reasons:

- The `categorical_columns.remove(['income', 'age')` line is a syntax error
- Even when adding a `]` or removing the `[`, the code is still invalid — instead, one must use two different calls to `remove`, one with `'income'` as the sole argument and one with `'age'` as the sole argument.
- The `disjoint_set` dependency, used in `mst.py`, is not installed automatically when running `pip install smartnoise-synth`, so the code snippet fails with `ModuleNotFoundError: No module named 'disjoint_set'`.
- After running `pip install disjoint-set`, it fails for a different import error: `ModuleNotFoundError: No module named 'networkx'`
- After running `pip install networkx`, it fails for a different reason: 

```
Please install mbi with:
   pip install git+https://github.com/ryan112358/private-pgm.git
Spent 0.5 epsilon on preprocessor, leaving 0.5 for training
Fitting with 367200 dimensions
Traceback (most recent call last):
  File "approx_bounds_synth.py", line 15, in <module>
    nullable=True
  File "/home/damien/.pyenv/versions/3.7.16/lib/python3.7/site-packages/snsynth/base.py", line 105, in fit_sample
    **kwargs
  File "/home/damien/.pyenv/versions/3.7.16/lib/python3.7/site-packages/snsynth/mst/mst.py", line 97, in fit
    domain = Domain(colnames, cards)
NameError: name 'Domain' is not defined
```

- At that point, running the suggested command (`pip install git+https://github.com/ryan112358/private-pgm.git`) fixes the error, and I can run the code snippet. But then, why is that "please run the following comment" simply *printed* to stdout, as opposed to thrown as an exception, so the end-user only sees this? Furthermore, why isn't it part of the automatically installed dependencies?

joshua-oss:
Thanks for reporting this. We will update the documentation and consider how to best alert people when they're trying to run a synthesizer that doesn't have the correct dependencies installed.

[...]

PR #581, opened 2023-11-04
Bump Release to 1.0.2

joshua-oss:
* Update snsql dependency to 1.0.2 to take advantage of bug fix in `approx_bounds`
* Fix bug where MST synthesizer would raise warning but attempt to keep running when `mbi` not installed
* Update documentation to highlight `mbi` dependency

Addresses #573


ocolluphid:
@joshua-oss question re. this update to the MST code: does this mean that the tutorial [here](https://github.com/opendp/smartnoise-sdk/blob/main/synth/samples/mst_sample/mst_sample_pums.ipynb) regarding domain use is no longer up to date and we should defer to this: https://docs.smartnoise.org/synth/synthesizers/mst.html

joshua-oss:
Yes, that code sample still should work (we haven't done anything to disable it) but it's better to use the common pattern used across all the synthesizers. Note that I recently discovered an issue where MST and AIM are failing if the inferred cliques overlap in a certain order, so if you're seeing errors like "column_name is both an index and a level", it's a bug that I haven't yet tracked down.

ocolluphid:
sounds good! mainly asking because it seems like the new style no longer requires domain specification and I'm guessing that's being automated in the background?

ocolluphid:
thanks! will defer to latest one without domain specification then.

Issue #576, opened 2023-10-20
Type hints


mhauru:
While trying to understand the smartnoise-sql code, I found that having Python type hints in place would have been useful. Given it's hacktoberfest, if I find the time, I might try adding some myself. It would be a good way to learn my way around the code too. @joshua-oss, would this be a welcome PR?

joshua-oss:
Definitely would be appreciated; thanks!

PR #584, opened 2023-11-17
More type hints

mhauru:
Started adding more type hints in a couple of files, and setup a pyright config so that if one runs pyright, those two files get checked.

In the process fixed something I think is a bug, see line 17 in `sql/reader/base.py`

Related to #576


mhauru:
I don't know what the GCP Action does, but the failure seems likely unrelated to this PR.

joshua-oss (approved):
Looks great; thanks for the contribution! The GCP failure is just an issue with the API key used to test against BigQuery.

PR #619, opened 2025-08-12
Bump version number for sn-eval fix

joshua-oss:
Update version so we can publish Paxton's eval fix to PyPi


paxton-coder (approved):
Thanks for the changes!

PR #623, opened 2026-02-04
feature: Update mbi to latest (1.0.0+) build for aim and mst.

tjh2822:
## Summary

Updates `mst.py` and `aim.py` to use the latest development build from MBI. The commit that smartnoise is using is about 3 years old and uses an outdated interface for the library (See https://github.com/ryan112358/mbi/issues/26). MBI has undergone significant development and this brings in the very latest changes. 

I intended to use the `v1.0.0` release of MBI from pypi but ran into a bug related `jax.jit`. This bug was fixed in https://github.com/ryan112358/mbi/pull/85 but a new release of MBI has yet to be put out. There were some minor changes between the API of `v1.0.0` and the latest development build (Dataset API). I don't know if there are bigger plans for a v1.1 of MBI but because of the jax issue this is the earliest commit that will work. 

Changes
* Updated MBI dependency
* Added MBI to pyproject.toml
* Updated documentation notes about MBI installation
* Replaced FactoredInterface (now missing) with `estimation.mirror_descent()`
* Replaced the tuple measurements (`Q`, `y`, `sigma`, `proj`) with new `LinearMeasurement`
* Replaced FactoredInterface with MarkovRandomField for aim. 
* Numpy arrays converted to jax arrays as needed

## Help Wanted
*  I did a little smoke test to test out the LinearMeasurement and I was happy with the results. I think it would be helpful for someone else to independently test the implementation to see if they get the same results. 
* We should do a new release of smartnoise. Not sure if we want to wait for a v1.1 of MBI or just release it now. Also, I think this might warrant a minor version bump but wanted to see what the maintainers were thinking.  
* I ran the tests `test_aim.py` and `test_mst.py` but there are only a couple of cases. Maybe we should add a more through test case? 

## Next Steps
* Release new version

### DIFF
--- a/synth/docs/source/synthesizers/aim.rst
+++ b/synth/docs/source/synthesizers/aim.rst
@@ -4,8 +4,6 @@ Adaptive Iterative Mechanism (AIM)
 
  `AIM <https://arxiv.org/abs/2201.12677>`_ is a workload-adaptive algorithm, within the paradigm of algorithms that first selects a set of queries, then privately measures those queries, and finally generates synthetic data from the noisy measurements. It uses a set of innovative features to iteratively select the most useful measurements, reflecting both their relevance to the workload and their value in approximating the input data. AIM consistently outperforms a wide variety of existing mechanisms across a variety of experimental settings.
 
-Before using AIM, install `Private-PGM <https://github.com/ryan112358/private-pgm.git>`_ :
-
 .. code-block:: bash
 
   pip install git+https://github.com/ryan112358/private-pgm.git
--- a/synth/snsynth/aim/aim.py
+++ b/synth/snsynth/aim/aim.py
@@ -1,32 +1,13 @@
 import numpy as np
 import pandas as pd
 
-try:
-    from mbi import FactoredInference, Dataset, Domain, GraphicalModel
-except ImportError:
-    print("Please install mbi with:\n   pip install git+https://github.com/ryan112358/private-pgm.git@01f02f17eba440f4e76c1d06fa5ee9eed0bd2bca")
-
+from mbi import Dataset, Domain, LinearMeasurement, estimation, MarkovRandomField, marginal_oracles
 import itertools
 from snsynth.base import Synthesizer
-from mbi import Dataset, FactoredInference, Domain
 from snsynth.utils import cdp_rho, exponential_mechanism, gaussian_noise, powerset
 from scipy import sparse
 
 prng = np.random
-
-
-class Identity(sparse.linalg.LinearOperator):
-    def __init__(self, n):
-        self.shape = (n,n)
-        self.dtype = np.float64
-    def _matmat(self, X):
-        return X
-    def __matmul__(self, X):
-        return X
-    def _transpose(self):
-        return self
-    def _adjoint(self):
-        return self
 
 def downward_closure(Ws):
     ans = set()
@@ -36,8 +17,9 @@ def downward_closure(Ws):
 
 
 def hypothetical_model_size(domain, cliques):
-    model = GraphicalModel(domain, cliques)
-    return model.size * 8 / 2 ** 20
+    """Estimate model size based on domain size and cliques"""
+    total_size = sum(domain.size(cl) for cl in cliques)
+    return total_size * 8 / 2 ** 20
 
 
 def compile_workload(workload):
@@ -69,7 +51,7 @@ class AIMSynthesizer(Synthesizer):
     :type verbose: bool
 
     Based on the code available in:
-    https://github.com/ryan112358/private-pgm/blob/master/mechanisms/aim.py
+    https://github.com/ryan112358/mbi/blob/master/mechanisms/aim.py
     """
 
     def __init__(self, epsilon=1., delta=1e-9, max_model_size=80, degree=2, num_marginals=None, max_cells: int = 10000,
@@ -135,8 +117,7 @@ class AIMSynthesizer(Synthesizer):
 
         self.rho = 0.0 if self.delta == 0.0 else cdp_rho(self.epsilon, self.delta)
 
-        data = pd.DataFrame(train_data, columns=colnames)
-        data = Dataset(df=data, domain=domain)
+        data = Dataset(pd.DataFrame(train_data, columns=colnames), domain=domain)
         workload = self.get_workload(
             data, degree=self.degree, max_cells=self.max_cells, num_marginals=self.num_marginals
         )
@@ -147,7 +128,11 @@ class AIMSynthesizer(Synthesizer):
         if samples is None:
             samples = self.num_rows
         data = self.synthesizer.synthetic_data(rows=samples)
-        data_iter = [tuple([c for c in t[1:]]) for t in data.df.itertuples()]
+
+        # need to reorder columns to match expected order
+        colnames = ["col" + str(i) for i in range(self._transformer.output_width)]
+        df = data.df[colnames]
+        data_iter = [tuple([c for c in t[1:]]) for t in df.itertuples()]
         return self._transformer.inverse_transform(data_iter)
 
     @staticmethod
@@ -179,6 +164,7 @@ class AIMSynthesizer(Synthesizer):
         # workload = [cl for cl, _ in W]
         candidates = compile_workload(workload)
         answers = {cl: data.project(cl).datavector() for cl in candidates}
+        total = float(len(data.df))
 
         oneway = [cl for cl in candidates if len(cl) == 1]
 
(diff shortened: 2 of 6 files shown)


synth/docs/source/synthesizers/aim.rst
```diff
@@ -4,8 +4,6 @@ Adaptive Iterative Mechanism (AIM)
 
  `AIM <https://arxiv.org/abs/2201.12677>`_ is a workload-adaptive algorithm, within the paradigm of algorithms that first selects a set of queries, then privately measures those queries, and finally generates synthetic data from the noisy measurements. It uses a set of innovative features to iteratively select the most useful measurements, reflecting both their relevance to the workload and their value in approximating the input data. AIM consistently outperforms a wide variety of existing mechanisms across a variety of experimental settings.
 
-Before using AIM, install `Private-PGM <https://github.com/ryan112358/private-pgm.git>`_ :
```
Shoeboxam, line 7: I notice that this PR makes MBI a mandatory dependency, whereas before it was optional. A decision for @joshua-oss.

synth/snsynth/aim/aim.py
```diff
@@ -1,33 +1,14 @@
 import numpy as np
 import pandas as pd
 
-try:
-    from mbi import FactoredInference, Dataset, Domain, GraphicalModel
-except ImportError:
-    print("Please install mbi with:\n   pip install git+https://github.com/ryan112358/private-pgm.git@01f02f17eba440f4e76c1d06fa5ee9eed0bd2bca")
-
+from mbi import Dataset, Domain, LinearMeasurement, estimation, MarkovRandomField, marginal_oracles
```
Shoeboxam, line 4: Nice to see this updated to the newer APIs. 👍

synth/snsynth/aim/aim.py
```diff
@@ -179,6 +164,7 @@ def AIM(self, data, workload):
         # workload = [cl for cl, _ in W]
         candidates = compile_workload(workload)
         answers = {cl: data.project(cl).datavector() for cl in candidates}
+        total = float(len(data.df))
```
Shoeboxam: I don't think this code previously (or should) make a bounded-DP assumption. So this would be an un-accounted epsilon=infty estimate of the total number of records in the dataset. The total argument in the mirror descent is optional, you could just leave it off (how it was before), thereby avoiding this issue.

synth/snsynth/aim/aim.py
```diff
@@ -189,13 +175,11 @@ def AIM(self, data, workload):
         print('Initial Sigma', sigma)
         rho_used = len(oneway) * 0.5 / sigma ** 2
         for cl in oneway:
-            x = data.project(cl).datavector()
-            y = x + gaussian_noise(sigma, x.size)
-            I = Identity(y.size)
-            measurements.append((I, y, sigma, cl))
+            x = np.asarray(data.project(cl).datavector())
+            y = x + np.asarray(gaussian_noise(sigma, x.size))
```
Shoeboxam, line 178: This is not your problem, but there are a few problems with the noise addition:

* `x + gaussian(0, sigma)` may technically be vulnerable to numerical issues affecting the privacy guarantee, whereas `gaussian(x, sigma)` is not. 
* Calling the scalar-valued mechanism `size` times incurs `size` times as much FFI overhead. 
* It's also not necessary to add float-valued noise; shouldn't x always be an integer array?

This takes these comments into account:
```python
# make the mechanism
dp.enable_features('contrib')
input_domain = dp.vector_domain(dp.atom_domain(T=int))
input_metric = dp.l2_distance(T=int)
meas = dp.m.make_gaussian(input_domain, input_metric, sigma)

# use the mechanism to make the release
y = meas(x)
```

synth/snsynth/aim/aim.py
```diff
@@ -189,13 +175,11 @@ def AIM(self, data, workload):
...
+            measurements.append(LinearMeasurement(y, cl, sigma))
 
-        engine = FactoredInference(data.domain, iters=1000, warm_start=True)
-        model = engine.estimate(measurements)
+        model = estimation.mirror_descent(data.domain, measurements, known_total=total, iters=1000, marginal_oracle=marginal_oracles.message_passing_stable)
```
Shoeboxam: Why change the default oracle?

synth/snsynth/aim/aim.py
```diff
@@ -147,7 +128,11 @@ def sample(self, samples=None):
         if samples is None:
             samples = self.num_rows
         data = self.synthesizer.synthetic_data(rows=samples)
-        data_iter = [tuple([c for c in t[1:]]) for t in data.df.itertuples()]
+
+        # need to reorder columns to match expected order
+        colnames = ["col" + str(i) for i in range(self._transformer.output_width)]
```
Shoeboxam, line 133: This seems like quite a specific set of names. Do the input column names always have this form?

synth/snsynth/aim/aim.py
```diff
@@ -215,13 +199,14 @@ def AIM(self, data, workload):
...
-            x = data.project(cl).datavector()
-            y = x + gaussian_noise(sigma, n)
-            measurements.append((Q, y, sigma, cl))
+
+            x = np.asarray(data.project(cl).datavector())
+            y = x + np.asarray(gaussian_noise(sigma, n))
             z = model.project(cl).datavector()
+            measurements.append(LinearMeasurement(y, cl, sigma))
+
+            model = estimation.mirror_descent(data.domain, measurements, known_total=total, iters=1000, marginal_oracle=marginal_oracles.message_passing_stable)
```
Shoeboxam: Similar comments.

synth/snsynth/mst/mst.py
```diff
@@ -140,28 +135,60 @@ def measure(self, data, cliques, sigma, weights=None):
...
+        #     supports[col] = sup
+        #     if supports[col].sum() == y.size:
+        #         new_measurements.append((Q, y, sigma, proj))
+        #     else:  # need to re-express measurement over the new domain
+        #         y2 = np.append(y[sup], y[~sup].sum())
+        #         I2 = np.ones(y2.size)
+        #         I2[-1] = 1.0 / np.sqrt(y.size - y2.size + 1.0)
+        #         y2[-1] /= np.sqrt(y.size - y2.size + 1.0)
+        #         I2 = sparse.diags(I2)
+        #         new_measurements.append((I2, y2, sigma, proj))
```
Shoeboxam: I think it would be better not to keep carrying dead code around. Why are you wanting to keep it around? Can you explain the change in behavior?

synth/snsynth/aim/aim.py
Shoeboxam: This could be handled as a separate issue if you'd rather, given that it was already this way in the code.

Shoeboxam (changes requested):
Thanks for opening the PR! Looks like you've also made some upstream changes to MBI too! That's great to see. 👍

synth/docs/source/synthesizers/aim.rst
tjh2822: I can revert this if needed. However, If we want to we could add an optional dependency group to the pyproject.toml for the marginal synthesizers?

synth/snsynth/mst/mst.py
tjh2822: This was an oversight, I've fixed it in the latest commit (ee4695f2f99a570a07277e2be97ac0716274e2ca).

synth/snsynth/aim/aim.py
tjh2822: The columns are hard coded this way in fit(). I also don’t like it and think it’s bug prone and might be a bug/oversight. I noticed it when I was debugging that the transformer expects the original order, not what the data frame puts out. MST seems to have a fix in the undo_compress_fn.

synth/snsynth/aim/aim.py
tjh2822: This was something that got lost when I moved away from FactorInterface. The old API was using the same thing under the hood but we have to manually specify it now.

synth/snsynth/aim/aim.py
tjh2822: Good catch, I've fixed this is commit 6bf6303fd9e94b9e80037d4317d22ffd4c52b445

synth/snsynth/aim/aim.py
tjh2822: I've tried to keep the code as close to the original as possible but this is a good call out. A separate issue will be best.

synth/docs/source/synthesizers/aim.rst
joshua-oss: I'm happy to have it as mandatory since we're updating to more recent code

synth/snsynth/aim/aim.py
joshua-oss: Agreed, we can make a separate issue

joshua-oss (approved):
Thanks for the submission! This was something that we had also run into that was causing us to pin to older code, so it will be great to be able to use the latest mbi.

joshua-oss:
> Thanks for the submission! This was something that we had also run into that was causing us to pin to older code, so it will be great to be able to use the latest mbi.

FYI, the CI is failing, but I think it's caused by the use of checkout@v2 which needs to be updated to @v4.  It works fine for me locally. I'm OK to merge this to main and then I can work out the CI issues. Will wait for @Shoeboxam confirmation that the requested change has been resolved to his satisfaction.

Shoeboxam (approved):
I see this now. Responses look good.

Issue #625, opened 2026-02-24
Compatibility update smartnoise-SQL for OpenDP 0.14.1


LancelotMarti:
Hello Smartnoise Team! Thanks again for the great work!

We are using a project using the opendp 0.14.1 version. Unfortunately, smartnoise-sql is currently only compatible with opendp < 0.13. Would it be possible to update smartnoise to make it compatible with the newer version of opendp?

Thank you very much for your help!

joshua-oss:
Thanks for the patch! The latest smartnoise packages on PyPi now support OpenDP 0.14.x

PR #626, opened 2026-03-05
Smartnoise-sql - Opendp 0.14.1: Fix float domains that implicitly allows NaN

LancelotMarti:
Pipelines using floating-point domains may fail during construction in v0.14 because `atom_domain(T=float)` implicitly allows `NaN` values. Transformations such as `then_clamp` and `then_sum` reject domains that may contain `NaN`, leading to the error:

`OpenDPException: FailedFunction("Domain contains nan")`

This change ensures pipelines behave as expected by explicitly disabling `NaN` in the input domain.


LancelotMarti:
Related to #625

joshua-oss (approved):
Thanks for the fix!

PR #637, opened 2026-04-15
Update to use latest MBI from PyPi

joshua-oss:
We couldn't publish to PyPi with pyproject linking to GitHub commit hash, but MBI has since been updated to 1.1.0 which is newer than the PR, so we just take the newer MBI now.


joshua-oss:
Flagging @tjh2822 for review. Everything is passing with your mbi PR and package is deployed to PyPi

